Issue #3, opened 2022-12-03
Search Animation


tisto:
When the user clicks on the search icon in the site header, an animation is triggered that reveals a search field.

Ideas:

- https://github.com/tobiasahlin/animated-menu
- https://hematogenix.com/ https://twitter.com/RitikaAgrawal08/status/1566429244321763328?s=20&t=W-TZFjWYQDNbmMD8lnabZw (CodePen: https://codepen.io/RitikaAgrawal08/full/oNdgzep)
- https://swiftype.com/

danalvrz:
Ready for review in our client project

PR #16, opened 2022-12-16
Header styling

danalvrz:
Regarding: https://github.com/kitconcept/bfs-intranet/issues/25
- Made independent wrappers for tools, and search in Header.jsx
- Styled tools and links
- Aligned logo, page links, and search icon
- Added animations to search icons
- Added animation to search-bar when active
- Added event listener to close search-bar with `Escape` key
- Added step to auto-select input text when search-bar is active
- Fixed responsiveness considering toolbar

Recording:

https://user-images.githubusercontent.com/89805481/208200498-014f8079-9c32-43fa-b310-4671d3bebe51.mov

### DIFF
--- a/src/theme/_header.scss
+++ b/src/theme/_header.scss
@@ -1,13 +1,3 @@
-// @mixin search($zIndex, $searchOn: false) {
-//   z-index: $zIndex;
-
-//   @if $searchOn {
-//     z-index: 11;
-//   } @else {
-//     z-index: auto;
-//   }
-// }
-
 .header-wrapper .header {
   display: flex;
   flex-direction: column-reverse;
@@ -24,7 +14,6 @@
       padding-left: 50px;
       z-index: 11;
     }
-    
 
     .navigation {
       flex-grow: 2;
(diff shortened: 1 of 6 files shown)


danalvrz:
I believe the z-index of the logo could be dynamically handled also, it is just an aesthetic decision inspired by the Hematogenix example given in  https://github.com/kitconcept/volto-light-theme/issues/3 . This change can be reversed, and omitted.

src/theme/_header.scss
```diff
@@ -90,25 +82,47 @@
...
+          transition-duration: 100ms;
         }
       }
 
       .search-bar {
         position: absolute;
+        padding-top: 2em;
         z-index: 10;
-        top: -152px;
+        top: -100%;
```
davisagli, line 115: @danalvrz I think this is 100% of the container height, not 100% of the search bar's height. It's probably the reason there is a slight delay before the search bar becomes visible when opening it. I would try setting this to -90px to match the height.

sneridagh:
LGTM so far! great work! We historically we are using rem's instead of em's. It's said that it's a better practice (at least the last time I checked 😅) and also Semantic UI encouraged it thorough its system.

The animation looks good to me too.

danalvrz:
@sneridagh yes, making the change to rem was the right call! Referencing the root element is more consistent

davisagli (approved):

PR #17, opened 2022-12-19
Footer styling

danalvrz:
Regarding: https://github.com/kitconcept/bfs-intranet/issues/26

- Resized footer to match Figma example
- Changed font-weights
- Aligned links to center and adjusted spacing
- Reduced font-size for "Powered by" link

Screenshot:
Screenshot

### DIFF
--- a/src/theme/_footer.scss
+++ b/src/theme/_footer.scss
@@ -8,11 +8,11 @@
     font-size: 18px;
     height: inherit;
 
-    > a:nth-child(4){
+    a.powered-by {
       font-size: 14px;
     }
 
-    div {
+    .footer-message {
       font-weight: $bold !important;
       a {
         font-weight: inherit;
--- a/src/theme/_footer.scss
+++ b/src/theme/_footer.scss
@@ -1,22 +1,37 @@
 #footer {
   text-align: center;
+  height: 429px;
 
   .footer {
-    padding: 5.5em 2em;
+    padding: 4.2em 2em;
     background-color: $lightgrey;
     font-size: 18px;
-    font-weight: $bold;
+    height: inherit;
+
+    > a:nth-child(4){
+      font-size: 14px;
+    }
+
+    div {
+      font-weight: $bold !important;
+      a {
+        font-weight: inherit;
+      }
+    }
+
 
     ul {
       display: flex;
       justify-content: center;
-      margin-top: 1.8em;
+      margin-top: 1.4em;
+      padding-left: 0;
       list-style: none;
 
       li {
-        padding: 0 5px;
+        padding: 0 7px;
         border-right: 1px solid #205c90;
         font-size: 18px;
+        
 
         &:last-of-type {
           border: none;
--- a/src/theme/_footer.scss
+++ b/src/theme/_footer.scss
@@ -1,19 +1,18 @@
 #footer {
   text-align: center;
-  height: 429px;
+  height: fit-content;
 
   .footer {
-    padding: 4.2em 2em;
+    padding: 4.2rem 2rem;
     background-color: $lightgrey;
     font-size: 18px;
-    height: inherit;
 
     a.powered-by {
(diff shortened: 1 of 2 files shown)


src/theme/_footer.scss
```diff
@@ -1,22 +1,37 @@
...
     background-color: $lightgrey;
     font-size: 18px;
-    font-weight: $bold;
+    height: inherit;
+
+    > a:nth-child(4){
+      font-size: 14px;
+    }
+
+    div {
```
sneridagh, line 15: As a rule, I never use "blind" selectors, as referring to bare div's or alike, if it's not absolutely necessary or doing otherwise is too much work/hassle.

src/theme/_footer.scss
```diff
@@ -1,22 +1,37 @@
...
 
   .footer {
-    padding: 5.5em 2em;
+    padding: 4.2em 2em;
     background-color: $lightgrey;
     font-size: 18px;
-    font-weight: $bold;
+    height: inherit;
+
+    > a:nth-child(4){
```
sneridagh, line 11: Why is the nth-child referring to? Can't we just drop a className on the thing?

sneridagh (commented):
LGTM, just a couple of comments

src/theme/_footer.scss
danalvrz: I added classes to both elements to avoid the blind selectors. Thanks!

src/theme/_footer.scss
```diff
@@ -1,22 +1,37 @@
 #footer {
   text-align: center;
+  height: 429px;
```
steffenri, line 3: @danalvrz Can we handle this differently? I would define the height of the footer via its elements + paddings on top and bottom. A fixed height like this we will have to change for mobile later, if we can do it more generic that would be great.

src/theme/_header.scss
```diff
@@ -123,14 +148,27 @@
             .searchbox {
               justify-content: space-between;
               border-left: none;
+              border-bottom: 2px solid $black;
+              margin-left: 12.6rem;
+              padding-bottom: 0.5rem;
+              padding-top: 0;
+
+              @media only screen and (max-width: $tablet-breakpoint) {
```
steffenri, line 156: @sneridagh @danalvrz So we have container queries now. Doesn't that make media queries obsolete? I wonder if it complicates things if we use both.

src/theme/_footer.scss
davisagli, line 3: I think it works fine already if we just take out the fixed height:

```suggestion
```

davisagli, line 3: I think it works fine already if we just take out the fixed height:

```suggestion
```

src/theme/_footer.scss
```diff
@@ -1,22 +1,37 @@
...
 
   .footer {
-    padding: 5.5em 2em;
+    padding: 4.2em 2em;
     background-color: $lightgrey;
     font-size: 18px;
-    font-weight: $bold;
+    height: inherit;
```
davisagli, line 9:
```suggestion
```

src/theme/_footer.scss
```diff
@@ -1,22 +1,37 @@
...
+
+    a.powered-by {
+      font-size: 14px;
+    }
+
+    .footer-message {
+      font-weight: $bold !important;
```
davisagli, line 16: It doesn't look like there's a reason it needs to be `!important`, so we should avoid using `!important`

src/theme/_header.scss
danalvrz: Great questions, I am not sure, but would be happy to find a different implementation if it does complicate things

davisagli: I think we will probably want to look at using container queries, but I would leave it until Victor is back and in a new PR.

davisagli (approved):

PR #30, opened 2023-01-05
Grid theme style

danalvrz:
Issue -> https://github.com/kitconcept/bfs-intranet/issues/16

To be merged along with: 
https://github.com/kitconcept/bfs-intranet/pull/59
https://github.com/kitconcept/volto-blocks-grid/pull/68


src/theme/blocks/_grid.scss
```diff
@@ -0,0 +1,98 @@
...
+      .grid-block-slate {
+        width: 32% !important;
+      }
+    }
+    .four {
+      .grid-block-slate {
+        width: 23.5% !important;
+      }
+    }
+  }
```
davisagli, line 97: @danalvrz I see why you are setting the widths like this (to make space for the gutter, since the slate block has its own padding that is unrelated to the grid columns). But I don't particularly like it. In DLR, we avoid this by adding another div wrapped around the slate block (https://github.com/kitconcept/dlr-internet/blob/main/frontend/src/addons/volto-dlr/src/index.js#L358). So we can keep the normal column widths, and then apply the background and padding on that additional wrapper.

It's probably worth checking with @sneridagh whether this is still the best way to go.

src/components/Blocks/Teaser/schema.js
```diff
@@ -29,10 +29,7 @@ export const teaserSchemaEnhancer = ({ schema, formData, intl }) => {
 
   defaultStylingSchema({ schema, formData, intl });
 
-  schema.properties.styles.schema.fieldsets[0].fields = [
-    'align',
-    'backgroundColor',
-  ];
+  schema.properties.styles.schema.fieldsets[0].fields = ['align'];
```
sneridagh, line 32: This is no longer needed since in @kitconcept/volto-blocks-grid the align styling button included in there. Maybe you have to merge the main branch again?

src/customizations/volto/components/manage/Blocks/StyleWrapper.jsx
```diff
@@ -0,0 +1,110 @@
+/**
```
sneridagh, line 1: I'd like to maintain this customisation out of the way until it's really necessary. I hoped to be able to get along without it, but could be that we need it anyways. 

@danalvrz why do you need it? which use case?

Also this helper was refactored for Volto 16, so it has to be adapted before including it.

sneridagh (changes requested):

src/customizations/volto/components/manage/Blocks/StyleWrapper.jsx
danalvrz: It was implemented to avoid having to set different widths to the columns in the grid, for example when adding a 2 column grid with 2 slate blocks next to each other.

Screenshot

 My first approach was to hard code widths (49% instead of 50%) that allow a gutter between them. The comment above is related to this point (https://github.com/kitconcept/volto-light-theme/pull/30#discussion_r1063621999) If there's a better way I can update it

davisagli: @danalvrz No -- that's what we used the ExtraAlignWrapper for, but I don't think we need the StyleWrapper.

danalvrz: Got it, I was confusing them. I will take a look and remove it then

davisagli (approved):
We still need to figure out what to do with the background colors, but I think it's a good step in the right direction.

davisagli:
@sneridagh ok to merge now?

sneridagh (approved):

PR #33, opened 2023-01-06
Teaser can be reseted from the sidebar

iFlameing:


davisagli:
@iFlameing This feature belongs in volto-blocks-grid, not in the theme. And @sneridagh already implemented it there: https://github.com/kitconcept/volto-blocks-grid/pull/64

For our client project, you just need to make sure that mrs.developer is configured to develop volto-blocks-grid from main

PR #35, opened 2023-01-09
Backport Search block from dlr

iFlameing:
Screenshot 2023-01-09 at 5 13 38 PM


src/components/Blocks/Search/TopSideFacets.jsx
```diff
@@ -0,0 +1,159 @@
+import React from 'react';
+import {
+  SearchInput,
+  SearchDetails,
+  Facets,
+  // FilterList,
+  SortOn,
+} from '@plone/volto/components/manage/Blocks/Search/components';
+import { Grid } from 'semantic-ui-react';
```
davisagli, line 9: @iFlameing For the volto-light-theme we are not using Semantic UI (see: https://github.com/kitconcept/volto-light-theme#it-should-not-use-any-semanticui-component-or-styling)

@sneridagh Can you make a suggestion how to handle the width of the facet columns instead?

src/theme/blocks/_search.scss
```diff
@@ -0,0 +1,159 @@
...
+#page-add,
+#page-edit {
+  .block.search {
+    .search-wrapper {
+      margin-bottom: 0px;
+    }
+
+    .search-input-resultscount-sort {
+      display: flex;
+      @media only screen and (max-width: 786px) {
```
davisagli, line 11: All the media queries should use the variables for the breakpoints, instead of hardcoding it

davisagli (changes requested):

src/components/Blocks/Search/TopSideFacets.jsx
steffenri: @iFlameing You can use the browser css grid option. That should require only some lines of css and a little bit more QA.

sneridagh: The Grid component is one of the basic structural components that we should provide out of any specific design system. There is a bunch of them that I have in mind. I will look to define the list of what we need the next week.

However, since we are dealing with a customisation from a component of Volto itself, and this is not going to change anytime soon, we should proceed with using the SemanticUI one. Once Volto makes the switch, we can update all the customizations. Good thing is that they are all controlled in the `customization` folder. I'll add it to the documentation.

src/theme/blocks/_search.scss
sneridagh: @iFlameing we don't have access to the breakpoints in the default theme here, but we should come up with our own. I'll add a ticket, so you can use them.

sneridagh (commented):
I left some comments.

https://github.com/kitconcept/volto-light-theme/issues/41

src/theme/blocks/_search.scss
sneridagh: https://github.com/kitconcept/volto-light-theme/issues/41

src/components/Blocks/Search/TopSideFacets.jsx
sneridagh: https://github.com/kitconcept/volto-light-theme/blob/main/README.md#volto-components-customizations-use-case

sneridagh:
@iFlameing please complete this with the comments.

iFlameing:
Started working on this.

src/theme/blocks/_search.scss
iFlameing: done

iFlameing:
@sneridagh ready for review.

PR #58, opened 2023-01-20
Fix alignment for hamburger icon

danalvrz:
- Added rule to justify navigation content to flex end when in mobile screen

### DIFF
--- a/src/theme/_header.scss
+++ b/src/theme/_header.scss
@@ -20,7 +20,7 @@
       flex-grow: 2;
       align-self: center;
       justify-content: center;
-      @media only screen and (max-width: $largest-mobile-screen) {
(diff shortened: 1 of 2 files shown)


src/theme/_header.scss
```diff
@@ -20,6 +20,9 @@
       flex-grow: 2;
       align-self: center;
       justify-content: center;
+      @media only screen and (max-width: $largest-mobile-screen) {
```
davisagli, line 23: @danalvrz The hamburger menu is also shown at tablet widths so I think this should be
```suggestion
      @media only screen and (max-width: $computer-width) {
```

davisagli (changes requested):

davisagli (approved):

PR #60, opened 2023-01-23
Add style to introduction block

danalvrz:
- Changed width to default-width
- Added font style

To be merged along with: 
https://github.com/kitconcept/volto-introduction-block/pull/1


davisagli (changes requested):
@danalvrz Looks good so far but some of the styles in the slate toolbar are not working in the introduction block:

- the bold and italic buttons do nothing
- we should probably remove the headline buttons within the introduction block
- the list buttons and blockquote button work, but then the text is not the right size for an introduction

These are not shown in the design, but you could see if it is easy to make them work better?

Screen Shot 2023-01-23 at 2 46 57 PM

Issue #63, opened 2023-01-27
Add style to Button block


iRohitSingh:
https://www.figma.com/file/eYUCbivnNrnjXjBDUizSVX/Plone-6?node-id=385%3A5812&t=ckjGqMUdSZvV5AKV-0

iRohitSingh:
Pr: https://github.com/kitconcept/volto-light-theme/pull/64

iFlameing:
I checked it locally and works perfectly fine.

@sneridagh closing this issue since it is done.

PR #64, opened 2023-01-27
Add style to Button block (#63)

iRohitSingh:


danalvrz (commented):
Looks great! I added some margins to have more space between the button and the paragraph

PR #70, opened 2023-02-16
PoC group dinamically blocks depending on its BGColor

sneridagh:
@davisagli I had this idea yesterday, I had to try it out. In case it's better and less error prone than the approach we used in DLR and maybe it allows less CSS covering all the use cases. We'll see.


sneridagh:
closing in favor of #83

PR #71, opened 2023-02-24
Add entry points to theme for allowing another add-on to customize it.

sneridagh:


sneridagh:
Merging, I need it right away.

davisagli:
@sneridagh Can we mention this in the readme?

PR #72, opened 2023-02-28
add mermaidBlock, table and other fixes

sneridagh:


davisagli (commented):
What is the mermaidBlock?

ichim-david:
> What is the mermaidBlock?
@davisagli since I've noticed a mermaid.js file load after using volto-light-theme I was curious myself to see the answer.
Turns out after a bit of digging that it's a block from volto-code-block
https://github.com/plonegovbr/volto-code-block/blob/main/src/components/Blocks/Mermaid/View.jsx

I'm leaving this comment here hoping it will be helpful for others if they search for mermaid and don't know what's that about.

davisagli:
More info about mermaid: https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

PR #76, opened 2023-04-21
WIP: Add font patterns PoC

danalvrz:


steffenri:
@danalvrz @sneridagh I really like this pattern thing and they way we approach styling with it.
My only problem with this is the naming. I think "first", "two" etc. is not descriptive enough. Could we maybe come up with better names?

tisto:
@steffenri @danalvrz FYI: the naming "one, two", "first, second" was just a placeholder since we did not want to concern ourselves with finding proper names. We still have to come up with proper names that make more sense.

sneridagh:
@danalvrz I like the pattern names that you chose! 

@tisto what do you think?

danalvrz:
@sneridagh I made some changes to test the mapping, and combined approaches (semantic/descriptive). We can edit it down, or make any adjustments needed.

sneridagh:
@danalvrz LGTM! I would want to know what's your opinion and the DX of this approach. Let's talk later.

src/theme/_typo-custom.scss
```diff
@@ -15,6 +15,10 @@ div {
   @include body-text();
 }
 
+strong > * {
```
davisagli, line 18: Is there a reason to do it for children of `strong` rather than `strong` itself?

src/theme/_typo-custom.scss
```diff
@@ -64,15 +66,40 @@ h2 {
...
+  padding-bottom: 10px;
+  letter-spacing: 1px;
+  text-transform: uppercase !important;
+}
+
+#page-add,
+#page-edit {
+  .head-title,
+  .head-title span {
+    user-select: none;
```
davisagli, line 100: Why are we disabling selection of the head-title?

src/theme/blocks/_button.scss
```diff
@@ -17,13 +17,12 @@
...
   margin-bottom: 1.5rem;
   background: none;
   border-radius: unset;
-  color: #000;
-  font-size: 16px;
-  font-weight: 900;
-  line-height: 20px;
+  color: $black;
+  padding: 8px 20px;
+  @include body-text-bold();
```
davisagli, line 27: It might be worth creating a mixin for all the button styles, since we also have buttons that are included in other blocks, etc.

src/theme/_typo-custom.scss
```diff
@@ -12,33 +12,37 @@ span,
 div {
```
davisagli, line 12: We might want to take div out of this list; it's a bit surprising for divs to have their own font styles assigned rather than inheriting from their container.

davisagli (commented):
In general I think this looks great and the use of mixins seems really nice for when we want to reuse the same styles with multiple selectors.

I made a few comments/questions but they're all pretty minor and non-blocking.

src/theme/_typo-custom.scss
danalvrz: When in Edit mode, the markup changes from just a `strong`/`em` to nested `span` elements, but I changed it to a less general approach.

src/theme/_typo-custom.scss
danalvrz: Since the `headtitle` value can only be modified from the sidebar, I thought it was odd to be able to set the cursor on it, but not edit the text. It can be removed, or we could find a different approach.

src/theme/_typo-custom.scss
davisagli: @danalvrz I see, it's needed then because we have font styles assigned to the generic `span` selector. My gut feeling is that we should remove font styles from `div` and `span` so that they inherit from their container, but let's see what @sneridagh thinks.

src/theme/_typo-custom.scss
davisagli: @danalvrz Okay, I see the logic there. Although it does have the downside of preventing the user from selecting the text to copy it to the clipboard, if they want to do that. No need to change anything for now.

tisto: Allowing the user to copy text to the clipboard is an important feature that we shouldn't prevent.

src/theme/_typo-custom.scss
sneridagh: for sure, we can't style divs alone.

PR #77, opened 2023-05-03
WIP: PoC Add abovetitle

danalvrz:
Note: depends on https://github.com/plone/volto/pull/4733

### DIFF
--- a/news/77.feature
+++ b/news/77.feature
@@ -1,0 +1,1 @@
+Add AboveTitle custom component to display Title header @danalvrz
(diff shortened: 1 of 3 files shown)


davisagli (commented):
This looks great to me, as long as https://github.com/plone/volto/pull/4733 gets merged.

@sneridagh If that one needs more discussion, then we could use a shadow here for now. But, if it is not time sensitive, maybe we can just wait.

sneridagh:
I would close this for now, since we have slots now, this one should be re-thinked. Please re-open if you think otherwise.

PR #81, opened 2023-05-24
Fix variables order, add `!default` to all theme vars

sneridagh:
Following: https://6.docs.plone.org/volto/addons/theme.html#variables-addonsthemecustomizationsvariables

Adding `!default` to all theme vars so they can be overridable from the add-ons. Move `addonsThemeCustomizationsVariables` to the top.


danalvrz (approved):
It all looks good!

PR #83, opened 2023-05-26
Add background grouping per background color

sneridagh:


davisagli (commented):
@sneridagh How can I test it?

One option rather than putting it here or in core would be a separate addon.

Any chance we could group them also in edit mode? I imagine that this could help achieve wysiwyg without needing to remember to duplicate CSS -- but it probably makes the drag and drop more complicated?

sneridagh:
> @sneridagh How can I test it?
> 
> One option rather than putting it here or in core would be a separate addon.

Indeed, that is an option... I'd like also to test drive it in full before it, let's see if we can find any limitation.

> Any chance we could group them also in edit mode? I imagine that this could help achieve wysiwyg without needing to remember to duplicate CSS -- but it probably makes the drag and drop more complicated?

That would be amazing indeed, but given the layers of wrappers (and most specially the one you hinted, the DnD component, which should have the same ancestors) inside don't know if that could be feasible... I will give it a try anyways.

PR #84, opened 2023-05-29
Add experimental support for automatic block grouping in the editor

sneridagh:
It seems feasible! however there is some UI/UX drawbacks:

image

The edit block wrapper gets "behind", and the drag and drop (the blocks could get huge, is cumbersome). Maybe this, combined with the new DnD that Rob has been working on and the Navigator...


sneridagh:
closing in favor of the Block Model v3 #330

PR #86, opened 2023-06-05
PoC custom ItemBodyTemplate

danalvrz:


sneridagh:
@danalvrz looks good!

Can you please use the pattern used in #88 ? So the customization points to a component in the package?

sneridagh:
@danalvrz wait for me to merge https://github.com/kitconcept/volto-light-theme/pull/88 and #87

sneridagh (approved):

src/components/Blocks/Listing/SummaryTemplate.jsx
```diff
@@ -0,0 +1,66 @@
...
+    link = <a href={href}>{linkTitle || href}</a>;
+  }
+
+  return (
+    <>
+      <div className="items">
+        {items.map((item) => {
+          const hasType = item['@type'];
+          const CustomItemBodyTemplate = config.getComponent({
+            name: 'ListingItemTemplate',
```
davisagli, line 29: @sneridagh @danalvrz Sorry I'm a bit late with my feedback. There are 2 changes I would submit for consideration:

1. I would call this component `SummaryListingItemTemplate` to make it clearer in the name that it is specifically for the summary variation and not other listing variations. (This one is not used for the default variation, nor the grid...)

2. It would be nice if we could register a default version of this template that is used if no component was registered for the specific content type. Then the default could be adjusted by shadowing just that template, rather than this one. But, I think this requires adding support for falling back to a default in the volto registry, and it might be too big a thing to consider.

danalvrz: I'll update the name as suggested in 1.

PR #88, opened 2023-06-06
Refactor customizations, add comments everywhere, customizations as a…

sneridagh:
… proxy to components in light theme to allow selective customizations in the theme not in Volto itself.

This is a (potentially, might not) breaking change.


src/customizations/volto/components/theme/Footer/Footer.jsx
```diff
@@ -1,135 +1,11 @@
...
- * @module components/theme/Footer/Footer
+ * OVERRIDE Footer.jsx
+ * REASON: This theme uses a custom pre-@plone/components component
+ * SemanticUI-free located at the components folder.
+ * To override it, override the @kitconcept/volto-light-theme one instead of
+ * this one.
  */
 
-import React from 'react';
+import Footer from '../../../../../components/Footer/Footer';
```
ericof, line 9: I will use '@kitconcept/volto-light-theme' instead of the `../../`

ericof (approved):
I like this approach. I would suggest you do not use `../../` to import components into customization, but the package name.

PR #90, opened 2023-06-07
Fix strong (bold) in view mode

sneridagh:


src/theme/_typo-custom.scss
```diff
@@ -8,7 +8,6 @@ h6,
 p,
 a,
 li,
-span,
```
davisagli, line 11: @sneridagh I think we can now remove the `strong span span` and `em span span` selectors below, which were there to override the font-weight from here.

sneridagh: uh, ok, let's do it using another approach

davisagli: No, you misunderstood me, your original change was good. We should not set any font styles for div and span, but let them inherit from their container. Otherwise we have to remember to set the font-weight for _every_ div and span.

sneridagh:
@davisagli how about now? I removed span and div as we talked about, generalize the strong, only setting the font-weight.

davisagli:
@sneridagh I think you need to push

davisagli:
This was fixed via a push to the main branch: https://github.com/kitconcept/volto-light-theme/commit/286ba8954d358c6867b4449c892a428f1864f771

PR #91, opened 2023-06-08
Grid listing variation

danalvrz:


sneridagh (approved):

src/components/Blocks/Listing/GridTemplate.jsx
```diff
@@ -0,0 +1,79 @@
...
+  } else if (href) {
+    link = <a href={href}>{linkTitle || href}</a>;
+  }
+
+  return (
+    <>
+      <div className="items">
+        {items.map((item) => {
+          const hasType = item['@type'];
+          const CustomItemBodyTemplate = config.getComponent({
```
sneridagh, line 28: Let's put this on the body of the component, here it falls a bit hidden.

sneridagh (changes requested):
I'd say, let's merge it after the change I just requested.

src/components/Blocks/Listing/GridTemplate.jsx
danalvrz: @sneridagh just to make sure that the change I made was what you meant, before merging

PR #102, opened 2023-06-16
Button margin update

danalvrz:
@sneridagh to take advantage of the margins collapsing

### DIFF
--- a/src/theme/_bgcolor-blocks-layout.scss
+++ b/src/theme/_bgcolor-blocks-layout.scss
@@ -17,6 +17,11 @@
   }
 }
 
+// If next block has different bgcolor, cancel all margin-bottom
+#page-document .block.next--has--different--backgroundColor {
+  margin-bottom: 0 !important;
+}
+
 // For grids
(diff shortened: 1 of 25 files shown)


src/theme/_bgcolor-blocks-layout.scss
```diff
@@ -18,24 +18,25 @@
 }
 
 // For grids
-
-.block.__grid {
+#page-document .blocks-group-wrapper .block.__grid {
```
iFlameing, line 21: I think using Id as selector should never used, because overriding make it worse. You have to add the id for overriding, except we should write css only using classname for exception only we should use id.

@sneridagh you will be better person to make this decision. This is just my general observation.

danalvrz: @iFlameing I agree with you, overriding a styling rule when it was defined using an id gets somewhat more complicated later on.

danalvrz:
Closing PR since we took a different approach and now use a mixin to set vertical spacing.

PR #104, opened 2023-06-19
Fix overlapping of text with toolbar and clear button of Grid image b…

iFlameing:
…lock

### DIFF
--- a/news/30.bugfix
+++ b/news/30.bugfix
@@ -1,0 +1,1 @@
+Fix overlapping of text with toolbar and clear button of Grid image block @iFlameing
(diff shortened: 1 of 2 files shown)


sneridagh:
@iFlameing I need to see this one, from now on, we need to be very careful and thoughtful with the layers order.

sneridagh (approved):

sneridagh:
@iFlameing could you please fix conflicts?

sneridagh:
Fixed in grid add-on, now we use the core grid, so it's not in place anymore.

PR #107, opened 2023-06-22
Fix CSS of Headline Blocks (#32)

iRohitSingh:
View : 

Screenshot 2023-06-22 at 14 53 33

Edit : 

Screenshot 2023-06-22 at 14 54 11

### DIFF
--- a/src/theme/blocks/_heading.scss
+++ b/src/theme/blocks/_heading.scss
@@ -1,0 +1,14 @@
+.block.heading{
+  margin-bottom: 0!important;
+  .heading-wrapper{
+    h2{
+      margin-top: 0;
+      margin-bottom: 0;
+      padding-top: 80px;
+      padding-bottom: 80px;
+      font-size: 33px;
+      line-height: 42px;
+
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


src/theme/blocks/_heading.scss
```diff
@@ -0,0 +1,13 @@
+.block.heading {
+  margin-bottom: 0 !important;
+  .heading-wrapper {
+    h2 {
+      padding-top: 80px;
```
sneridagh, line 5: We have mixins for this now in `variables`. What was the issue? I thought this was fixed.

iRohitSingh: Edit mode is not similiar to view mode.
screenshot
Screenshot 2023-06-23 at 14 24 05

iRohitSingh: I have fix the problem using padding making similiar to dlr.

sneridagh:
@iRohitSingh problem is that we are using a completely different approach with the light theme and the auto grouping.
The fix you are searching only have to work on edit, then only apply there. If you take a closer look, in view mode, the spacing is broken now (your screenshot), because you add the padding above the existing margins.

sneridagh:
@iRohitSingh no worries, let me take care of it!

sneridagh:
@iRohitSingh I just tried and it's fine as it is, true that it does not forces the margins in the same way as in DLR, but I feel that it's better like this.

image

image

PR #116, opened 2023-06-23
Fix css issue of image block full width variante

### LINKED ISSUE #115: Fix css issue of image block full width variante

iFlameing:
Fixes #115

### DIFF
--- a/news/115.bugfix
+++ b/news/115.bugfix
@@ -1,0 +1,1 @@
+Fix css issue of image block full width variante @iFlameing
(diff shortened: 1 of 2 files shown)


danalvrz (approved):
LGTM, just a couple questions. Why is the breakpoint `1470px` specifically? And where can I find the design spec for this change?

iFlameing:
@danalvrz We have a full-width snap at 1440px. I have added another 30px because that is the currently available padding-left for full width. That's why I added another 30px to make it looks nice and consistent.

PR #117, opened 2023-06-23
Fix border for last Teaser

danalvrz:
To comply with the Vertical Spacing rule for the last Teaser


danalvrz:
@davisagli I noticed that as well, and believe it is because some classes are not present in edit mode, for example `next--is--[block type]`.  If needed, we can look into making that change.

davisagli:
Right, it depends on the classes added by the block wrapper, and we don't have that in edit mode. Okay for now.

sneridagh (approved):

PR #124, opened 2023-07-03
Fix Event content type (#16)

iRohitSingh:
My-test-Event (1)

PR #126, opened 2023-07-04
News Item view

iFlameing:


sneridagh:
@iFlameing how this is related with #124? 

Also please use `FormattedDate` component from core.

iFlameing:
code analysis is falling because I can't run the make format individually in this repo. I tried to develop in bfs is something is wrong with my setup.

iFlameing:
only format is wrong. Everything else is right.

PR #142, opened 2023-07-12
Sneridagh last 12 07

sneridagh:


danalvrz (approved):
looks good, the buttons on the sides of the block editor are in the right place now

PR #143, opened 2023-07-12
Fix Accordion CSS & other minor style fixes

danalvrz:


sneridagh (approved):

src/theme/blocks/_accordion.scss
```diff
@@ -2,3 +2,54 @@
...
+        color: $darkGrey;
+      }
+
+      &.active > span {
+        color: $black;
+      }
+
+      &.active > svg {
+        fill: $black !important;
+        transform: rotate(180deg);
```
sneridagh, line 34: @danalvrz I think they can be configured in the block's config... it's not documented, but take a look. (Maybe it's not worth now...)

PR #165, opened 2023-07-19
Fix image view in edit mode.

iFlameing:


src/customizations/volto/components/theme/View/ImageView.jsx
```diff
@@ -0,0 +1,3 @@
+import ImageView from '../../../../../components/Theme/ImageView';
```
sneridagh, line 1: Reason :) !!

sneridagh (changes requested):

sneridagh (approved):

PR #171, opened 2023-07-20
Fix responsive of heading block and heading of content type

iRohitSingh:


sneridagh (commented):
@iRohitSingh I'd like that all font related changes is done by Dante, who took care of the initial ones. So the work is cohesive and grouped. 

We can't have a generic mixing for this... might be that there are several font sizes with different margins.

@danalvrz could you please take care of this?

PR #177, opened 2023-07-24
Update to Volto 17a21 + Deprecate `volto-image-block` - Use new core Image component

sneridagh:
We have to be careful with this one, but this is the initial work needed.

The gist: I think the DLR image component is overengineered and not usable, and the DX is very poor. I still cannot make it work and instantiate it at the first shot. I think we should move to use the new image component in core. (Volto 17a21). The DX and the simplicity of it is far better.

The only downside is that the core one does not use `blurhash`. Any other opinions?

**NOTE for this August (while I'm on holidays)**: You can update Volto to latest without this, the image component in `volto-image-block` will take over. Up to you to merge this and follow up, an exhaustive QA should be done on all places where this is used.

### DIFF
--- a/src/components/Blocks/Image/ImageSidebar.jsx
+++ b/src/components/Blocks/Image/ImageSidebar.jsx
@@ -28,13 +28,15 @@ const ImageSidebar = (props) => {
             basic
             disabled={!data.url}
             onClick={() => {
-              onChangeBlock(block, {
-                ...data,
-                url: undefined,
-                image_scales: undefined,
-                image_field: undefined,
-                alt: data.url.title === data.alt ? undefined : data.alt,
+              // START CUSTOMIZATION
+              dataAdapter({
+                block,
+                data,
+                id: 'url',
+                onChangeBlock,
+                value: null,
               });
+              // END CUSTOMIZATION
             }}
           >
             <Icon name={trashSVG} size="24px" color="red" />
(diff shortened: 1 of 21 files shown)


ericof:
@sneridagh Maybe you need to re-run yarn?

ericof (approved):

src/components/Blocks/Image/ImageSidebar.jsx
```diff
@@ -0,0 +1,128 @@
...
+        onChangeField={(id, value) => {
+          dataAdapter({
+            block,
+            data,
+            id,
+            onChangeBlock,
+            value,
+          });
+        }}
+        // END CUSTOMIZATION
```
davisagli, line 101: @sneridagh This customization should go into volto, right?

src/components/Theme/ImageView.jsx
```diff
@@ -25,13 +25,20 @@ const ImageView = ({ content }) => {
...
   return (
     <Container id="page-document" className="view-wrapper image-view">
       {/* BEGIN CUSTOMIZATION */}
       <h1 className="documentFirstHeading">{content.title}</h1>
       {content?.image?.download && (
         <figure>
           <Image
+            item={content}
+            imageField="image"
+            alt={content.title}
```
davisagli, line 38: We shouldn't reuse the title as the `alt` text here, since it also appears in the caption, so would be read twice by screen readers.

PR #179, opened 2023-07-25
Fix responsiveness of video block

iFlameing:


src/theme/blocks/_video.scss
```diff
@@ -0,0 +1,13 @@
+.block.video {
+  &.align.left,
+  &.align.right,
+  &.align.center {
+    .video-inner {
+      @media only screen and (max-width: $tablet-breakpoint) {
```
iFlameing, line 6: do you have an idea to do it differently?

src/theme/blocks/_video.scss
```diff
@@ -0,0 +1,4 @@
+.block.video.align.right .video-inner,
+.block.video.align.left .video-inner {
+  @include overrideWidth();
```
sneridagh, line 3: I think it's too premature to use a mixin here, just for this use case... if there's more, then we can create it.

sneridagh (changes requested):

src/theme/blocks/_video.scss
iFlameing: done.

sneridagh (approved):

PR #181, opened 2023-07-26
fix responsive spacings & font sizes

danalvrz:
- Disabled dynamic fonts for in-grid headings in mobile
- Fixed top/bottom paddings for grid columns
- Added responsive media queries to spacing mixins
- Other minor spacing bugfixes

### DIFF
--- a/src/theme/_bgcolor-blocks-layout.scss
+++ b/src/theme/_bgcolor-blocks-layout.scss
@@ -1,6 +1,5 @@
 .blocks-group-wrapper {
-  padding-top: $color-block-change-vertical-spacing;
-  padding-bottom: $color-block-change-vertical-spacing;
+  @include color-block-change-vertical-spacing();
 }
 
 .blocks-group-wrapper.grey {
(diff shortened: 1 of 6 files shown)


sneridagh (approved):

sneridagh:
@danalvrz @iFlameing 

I've been thinking on the subject on when to use media queries and when to use container queries. It has to do with the thing that we were discussing yesterday about the edit mode 1:1 with the view mode. So I came up with a rule (which we can discuss, of course):

We use media queries when the styling it's generic enough to apply only to the View. 

We use container queries when do care explicitly about how the styling is being applied in edit mode as well and we want the content area to behave 1:1 with the view mode.

Reason: The container queries allow us to abstract the width from the sidebar and toolbar in edit mode, showing the content area as it will be in that size, in view mode.

Remember: The margins in responsive are being taken care with container queries in `layout.scss`. So everything related to that, goes like it works in there, with container queries. See implementations for details in case you need it.

I will put this info in the README.

I'm merging this one, since I see that all changes applies to media queries. If we need to refine them because we see that the edit mode differs too much, we can move them to container queries. It seems to me that all are related to spacing which does not make much sense in edit mode, and font sizes, which is also ok in edit mode (for now).

PR #182, opened 2023-07-27
Fix responsive spacings and heading font sizes

danalvrz:


src/theme/blocks/_grid.scss
```diff
@@ -71,6 +71,8 @@
     .teaser-item.default {
       padding-bottom: 0;
       border: none;
+      margin-right: auto !important;
```
iFlameing, line 74: can we somehow avoid the important parameter. Or it is something needed for semantic ui.

davisagli (approved):
LGTM so far

PR #37, opened 2023-01-11
Add proper testing ci and changelog via towncrier

sneridagh:

PR #187, opened 2023-07-31
Fix slider Block (#37)

iRohitSingh:


iRohitSingh:
@davisagli  I am running make i18n but it is not updating the translation. Also please review it. From my side I think it is working perfectly fine.

Screenshot 2023-08-02 at 20 45 14

src/theme/_variables.scss
```diff
@@ -32,6 +32,7 @@ $header-font: var(--custom-headers-font, $page-font) !default;
 // Mobile / Tablet portrait < 769 | Tablet landscape / Small desktop < 941 | Computer desktop < 1440 | Large Monitor > 1440
 $largest-mobile-screen: 768px !default; // not finished in `-width` for historical reasons (SemanticUI naming)
 $tablet-breakpoint: 769px !default;
+$largest-tablet-breakpoint: 1024px !default;
```
davisagli, line 35: Where did 1024 come from? We can't just make up a new breakpoint; they are part of the design. I think the places where you are using $largest-tablet-breakpoint below should use $computer-width instead.

davisagli:
@iRohitSingh About the i18n...I need to fix the i18n script in volto so that it can work when it is running inside a docker container in a different folder from the addon. I didn't have time to do that yet, so I used a workaround to run the script, and added a commit with the result.

davisagli (changes requested):
There are still a few things that need work here. I'm not going to prioritize it today, since this block isn't enabled for our client project.

iRohitSingh:
@davisagli Example content of slider block https://github.com/kitconcept/kitconcept.voltolighttheme/pull/11

And also check CSS it 100% like figma.

I don't understand fix container width and I don't know how to configure image component. I also fixed the case of empty slider block.

davisagli:
@iRohitSingh Thanks. I will look at the other todo items.

src/theme/blocks/_slider.scss
```diff
@@ -74,62 +72,63 @@
...
+    .title {
+      .supertitle {
+        //    margin-bottom: 20px !important;
+        font-size: 14px;
+        font-weight: 700;
+        letter-spacing: 1px;
+        line-height: 18px;
+        text-transform: uppercase;
+      }
+      h2 {
```
danalvrz, line 100: Here I recommend using the mixin `text-heading-h2()` , and remove the top margin to match better the implementation for other headings.

src/theme/blocks/_slider.scss
```diff
@@ -74,62 +72,63 @@
...
 
-    .supertitle {
-      color: $white;
-      font-family: 'Inter';
-      font-size: 20px;
-      font-weight: 700;
-      letter-spacing: 1px;
-      line-height: 24px;
+    .title {
+      .supertitle {
```
danalvrz, line 92: I would add the 20px of space as a bottom margin for the `supertitle`

src/theme/blocks/_slider.scss
```diff
@@ -9,34 +9,31 @@
...
-    color: #fff;
-    font-size: 48px;
-    font-style: normal;
-    font-weight: 700;
-    line-height: 56px;
-
-    @media only screen and (max-width: $largest-mobile-screen) {
-      @include word-break;
-    }
+  .slick-arrow {
```
danalvrz, line 20: According to the design, the arrows should only be visible for desktop when hovered over. I would recommend making the `.slick-arrow` element same height as the slide to make the hover action easier (should be triggered at the ends of the slide at any height). We could have `.slick-prev` and `.slick-next` nested in `.slick-arrow`  with the `&` operator since they will share most of the styles.

src/theme/blocks/_slider.scss
```diff
@@ -9,34 +9,31 @@
   .grid-teaser-item {
     flex-direction: column;
   }
+  .teaser-item {
```
danalvrz, line 12: It is a small detail, but there is a small segment of screen sizes (I think 941px - 1100px) that breaks the layout a little because the `.teaser-item` is bigger than the slide:
Screenshot 2023-08-04 at 14 04 20

danalvrz (changes requested):
It looks really nice! I just wanted to mention a few small details, please take a look at the comments. We can discuss any of them further.
// cc @davisagli

src/theme/blocks/_slider.scss
iRohitSingh: fixed

src/theme/blocks/_slider.scss
iRohitSingh: fixed

src/theme/blocks/_slider.scss
iRohitSingh: @danalvrz It is not easy to implement. I tried but didn't find correct solution for it. First I can't make the button full height and 50% width because we have continue reading button to see the content. Also I don't have control over html to try some stuff. The only solution will be using javascript which is also not easy. If you have any idea please let me know.

src/theme/blocks/_slider.scss
iRohitSingh: @davisagli @danalvrz This is same problem with landascape mode. For that I have written the 1041px style specially for this block in dlr. I also tried the same thing but david told me that we need figma for that. 
@steffenri your feedback is also needed because we may need design for this by thomas.

sneridagh:
@iRohitSingh what's the status of this one? What is left to do? Should I jump somewhere?

iRohitSingh:
@sneridagh use img srcset, and fixed container. @davisagli told me he will take care the remaining issue. Still need to some time for implementation.

src/theme/blocks/_slider.scss
sneridagh: I fixed it reducing the paddings from <1024px (with a container query).

Issue #22, opened 2022-12-22
Img component not updating if used in the sidebar


sneridagh:
While working in an improvement to the Teaser block (discarded), we would have a preview of the original object in the sidebar.
See video:

https://user-images.githubusercontent.com/486927/209126368-d175c145-7e0d-4cb8-b503-9d3c4c853d95.mov

It seems like and edge case, but worth taking a look at it.

sneridagh:
Closed, we are using Volto's default image component

PR #188, opened 2023-08-01
Add external link for slate block(#22)

iRohitSingh:
Screenshot 2023-08-01 at 16 48 26

### DIFF
--- a/news/188.bugfix
+++ b/news/188.bugfix
@@ -1,0 +1,1 @@
+Add external link for slate block @iRohitSingh
--- a/src/theme/_content.scss
+++ b/src/theme/_content.scss
@@ -194,7 +194,6 @@
   aspect-ratio: $aspect-ratio;
 }
 
-
 // Slate Style External Links
 .content-area .ui.container,
 .content-area .a.container {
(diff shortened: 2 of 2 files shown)


iFlameing (approved):
LGTM!

danalvrz:
> LGTM!

@iFlameing if possible, please let me know about changes before merging so at least a couple people get the chance to take a look.

davisagli:
@iFlameing The problem with this solution (using the CSS selectors based on the `href` attribute) is that we have to keep reimplementing the same CSS for each project, with a different URL.

Instead, we could add a feature to the UniversalLink component in volto to add an "external" class to external links. Then we could include CSS in volto-light-theme for `.external` and it would be the same CSS for every project. UniversalLink already checks if the link is external, so it should not be hard to implement. What do you think?

/cc @sneridagh @danalvrz

sneridagh:
@iFlameing @iRohitSingh @danalvrz @davisagli +1 for the implementation suggested by David. Still catching up, maybe you already implemented it!

iFlameing:
@sneridagh Doing it right now.

PR #194, opened 2023-08-02
Fix long title not working

iFlameing:


src/theme/_typo-custom.scss
```diff
@@ -65,6 +77,7 @@ h1.documentFirstHeading {
 // Block Title
 .block h2.headline {
   @include block-title();
+  @include word-break();
```
davisagli, line 80: We don't need this for specific selectors like this one, because it inherits from the general h2 selector below. It just adds unnecessary bytes to the output. I'll try to clean it up a bit...

davisagli (approved):

PR #196, opened 2023-08-03
Fix image rights

iFlameing:


src/theme/_layout.scss
```diff
@@ -97,7 +97,8 @@ footer {
 }
 
 // Container definitions for elements
-#page-document > * {
+#page-document > *,
+#page-add > * {
```
davisagli, line 101:
```suggestion
#page-document > * {
```

As far as I can see, this isn't needed. There are already CSS rules for block width on the add page. If you found a problem, please show it with a link and screenshot in the issue so that I can understand.

news/197.bugfix
```diff
@@ -0,0 +1 @@
+Fix image rights and add view not working properly @iFlameing.
```
davisagli, line 1:
```suggestion
Fix image rights @iFlameing.
```

davisagli (approved):

PR #199, opened 2023-08-03
Fix aspect-ratio and add general rule to center object position for images

### LINKED ISSUE #198: object-position centered as a general rule for all images
The rule `object-position: center;` is needed in a few places, it maybe would be worth setting it at a general level.

danalvrz:
fixes #198


davisagli (approved):

sneridagh:
@davisagli @danalvrz maybe we should document this, since it could be an opinionated decision.

PR #207, opened 2023-08-08
overwrite icons, default image & minor sizing fixes

danalvrz:
- Overwrite sorting icons
- Overwrite default image for Listings
- Fix image fit for Listings
- Remove duplicated styles in Search block
- Fix facets width 
- Add `searchBlock-container` class


src/theme/blocks/_listing.scss
```diff
@@ -250,6 +251,12 @@
   }
 }
 
+// To overwrite semantic ui .grid negative margin rule
+.ui.grid {
+  margin-right: 0;
+  margin-left: 0;
+}
```
davisagli, line 258: @danalvrz Tell me more about where this is solving a problem. At first glance it looks way too general for me to be confident it won't cause problems elsewhere.

src/theme/blocks/_search.scss
```diff
@@ -24,14 +24,12 @@
...
-          width: 220px;
-          height: min-content;
-          aspect-ratio: $aspect-ratio;
-          @container (max-width: #{$largest-mobile-screen}) {
-            width: 100%;
-          }
+      img {
+        width: 220px;
+        height: min-content;
+        aspect-ratio: $aspect-ratio;
```
davisagli, line 30: @danalvrz I think we should add `object-fit: cover` here, to make sure the default image fills the space when the project is using a different aspect ratio.

Screenshots from our project using 4/3... Without it:
Screenshot 2023-08-08 at 12 30 57 PM

With `object-fit: cover`:
Screenshot 2023-08-08 at 12 31 03 PM

src/theme/blocks/_search.scss
```diff
@@ -24,14 +24,12 @@
...
+        @media only screen and (max-width: $largest-mobile-screen) {
```
davisagli, line 31: @danalvrz Why'd you change it to use a media query instead of a container query?

davisagli (changes requested):

src/theme/blocks/_listing.scss
danalvrz: I moved the rule inside the Search block, I think it is better. This is the issue generated by the rule coming from semantic ui:
Screenshot 2023-08-08 at 14 10 32

src/theme/blocks/_search.scss
danalvrz: I added the `object-fit: cover` to all images inside a Listing, which applies to the Search results as well

src/theme/blocks/_listing.scss
davisagli: @danalvrz Got it. Looks safer now

src/theme/blocks/_search.scss
danalvrz: When in Tablet view, if you open the sidebar the container query triggers the `width: 100%`, breaking the layout for the images. If it is a media query, the screen size remains the same and the width change is not triggered:
Screenshot 2023-08-08 at 14 24 16

src/theme/blocks/_search.scss
davisagli: @danalvrz Okay. How come the search block has some unique image styles here? In general listings in the search block are supposed to look the same as the listing block.

src/theme/blocks/_search.scss
davisagli: @danalvrz Well, the point of the container query is to style things based on the actual size of the content area (what is left without counting the width of the sidebars). If that width is less than the mobile breakpoint width, then I would expect to see the mobile styles, even if the viewport width is still within the tablet range.

Here, it looks like the actual problem is that the mobile width inside the container is not styled to match the design (images are not supposed to be 100% width there):
Screenshot 2023-08-08 at 1 38 48 PM

https://www.figma.com/file/eYUCbivnNrnjXjBDUizSVX/Volto-Light-Theme?type=design&node-id=4069-35606&mode=dev

danalvrz: good point, let's leave it as a container query. We also should fix the order of the elements in mobile ( heading -> image -> description). I'll work on it.

src/theme/blocks/_search.scss
danalvrz: I deleted the styles for the images inside the Search block, they are not needed, and were basically duplicates of what we have for images in the Listings

src/theme/blocks/_search.scss
danalvrz: Fixed to match updated design: https://www.figma.com/file/eYUCbivnNrnjXjBDUizSVX/Volto-Light-Theme?node-id=4773%3A65840&mode=dev

Screenshot 2023-08-11 at 15 19 40

src/theme/blocks/_search.scss
danalvrz: Example project using 4/3:
Screenshot 2023-08-11 at 15-25-49 Block Auflistung (Listing)

src/theme/_layout.scss
```diff
@@ -172,7 +172,10 @@ footer {
   & > .block.listing .image-gallery,
   & > .block.listing.grid,
   & > .block.search .listing-item,
+  & > .block.search.grid,
   & > .block.search .search-input-resultscount-sort,
+  & > .block.search .facets,
+  & > .block.search .emptyListing,
```
davisagli, line 178: @danalvrz This is okay, but it makes me think that maybe we should put one container inside the grid block markup so that we don't need so many different CSS rules.

src/theme/blocks/_search.scss
```diff
@@ -28,20 +28,15 @@
         color: $black;
         @include text-heading-h2();
       }
-      &:last-child {
+      &:last-child:not(.facets) {
```
davisagli, line 31: This will expand out to some weird CSS like: `.facets:last-child:not(.facets)`. It would be better to repeat `.listing.message` and `.listing-item` in the scss so that the CSS rules are added for them only.

src/theme/blocks/_listing.scss
```diff
@@ -248,6 +255,12 @@
       text-align: center;
     }
   }
+
+  // To overwrite semantic ui .grid negative margin rule
+  .ui.grid {
+    margin-right: 0;
+    margin-left: 0;
+  }
```
davisagli, line 263: @danalvrz It looks like this is causing a misalignment of the facets

Screenshot 2023-08-15 at 9 55 02 AM

src/theme/_layout.scss
danalvrz: I agree that it would be better to have a container for it. I'll make a ticket about it.

davisagli: @danalvrz Perfect. It doesn't need to be solved in this PR as long as we have a ticket to come back to

news/207.bugfix
```diff
@@ -0,0 +1 @@
+Add icons and default image customizations, fix other small image sizing issues for Listings. @danalvrz
```
davisagli, line 1:
```suggestion
Add icons and default image customizations, fix other small image sizing issues for Listings, fix spacing issues for Search block. @danalvrz
```

davisagli (approved):

PR #215, opened 2023-08-11
Fix Translations for the accordeon block(#53)

iRohitSingh:


news/215.bugfix
```diff
@@ -0,0 +1 @@
+Fix Translations for the accordeon block @iRohitSingh
```
davisagli, line 1:
```suggestion
Fix Translations for the accordion block @iRohitSingh
```

davisagli (changes requested):
This should be fixed in https://github.com/eea/volto-accordion-block rather than in volto-light-theme

PR #216, opened 2023-08-11
Table block review

iFlameing:
table-block-comparison


src/theme/blocks/_table.scss
```diff
@@ -5,8 +5,14 @@
...
     p {
       font-weight: bold;
+      text-transform: none;
     }
   }
+
+  tr td {
+    border-top: 1px solid #c1c1c1;
+    border-left: 1px solid #c1c1c1;
+  }
```
danalvrz, line 15: To better match the design, it is necessary for the border to be 2px. It is also needed for the table, and the `th` elements.
Screenshot 2023-08-15 at 14 05 24

danalvrz (changes requested):
It would be better for usability, if the table has `overflow-x:auto` for mobile. Currently the layout gets affected by some media queries coming from semantic UI. The idea would be for the table not to change width nor layout, and allow horizontal scroll in smaller screens.

steffenri:
> It would be better for usability, if the table has `overflow-x:auto` for mobile. Currently the layout gets affected by some media queries coming from semantic UI. The idea would be for the table not to change width nor layout, and allow horizontal scroll in smaller screens.

@danalvrz Tables are an unsolved mistery on mobile devices. If we go with `overflow-x:auto` then basically every page that has a table (and for bfs, there are a lot) would have white space on the left/right and scrolling down the page would show that. In my opinion this would look terrible (we could test it though). 

I always dreamt about an option for mobile that you can click the table and have a kind of view that only shows the table. Then you can scroll, zoom, whatever you like. I think there are few options to really optimize tables for mobile.
We should discuss this anytime in the future, but for now (and for BfS) I would spend more time on this. It's just to complicated.

davisagli:
> @danalvrz Tables are an unsolved mistery on mobile devices. If we go with overflow-x:auto then basically every page that has a table (and for bfs, there are a lot) would have white space on the left/right and scrolling down the page would show that. In my opinion this would look terrible (we could test it though).

@steffenri You could set the table's min-width (instead of width) to the default container width if you want to avoid whitespace but still allow scrolling to see a table if its content is wider.

> I always dreamt about an option for mobile that you can click the table and have a kind of view that only shows the table. Then you can scroll, zoom, whatever you like. I think there are few options to really optimize tables for mobile.

Yeah, that could work.

> for now (and for BfS) I wouldn't spend more time on this. It's just to complicated.

I agree for BfS. This came up in the context of VLT yesterday and I suggested overflow-x to Dante as a quick improvement. But we can certainly leave it alone for now and come back to it later.

danalvrz:
@davisagli @steffenri @iFlameing then if it sounds good to you, we can close the block ticket once the desktop version is ready, and open a new one just for the mobile screens that we can revisit when makes the most sense.

davisagli:
@danalvrz Sounds good to me

davisagli:
Well, what I said won't work well. Setting a min-width of the default container width will make some tables too wide on mobile viewports (it will force them to have a scrollable area even if the content fits inside the viewport)

davisagli:
But maybe: min-width of default-container-width above a certain breakpoint, and min-width of 100% below it.

iFlameing:
@danalvrz I have double-checked the border and in Figma it is 1px. Attaching a screenshot for your reference. If we still have doubts we can talk today after wrap-up. thank you for having a detailed conversation about how we are going to show the table in mobile view. I completely want to re-implement a new view for the table field in mobile which shows only necessary things with pagination.  You click on pagination and then you will be able to see more. But I don't want to scroll x to see other data and lose context of what I want to see in the table.

But it is very complicated becasue of its dynamic nature .

Screenshot 2023-08-17 at 1 27 29 PM

iFlameing:
@danalvrz Thomas bork said that we should stick to 1px solid border. Please review it once again.

sneridagh:
@davisagli I have the same doubt... don't know if we could spend some time to remove any traces of SemanticUI Table in the Table block.

/cc @tisto

PR #219, opened 2023-08-12
Add settings.showSelfRegistration check to Header override

fredvd:
to correctly show/hide 'register' menu item.


davisagli (commented):
@fredvd I think it should go inside the Anontools component so it's only shown to users who haven't already logged in.

davisagli (approved):

PR #239, opened 2023-09-26
remove huge margins on listing blocks without headlines.

steffenri:
example: https://light-theme.kitconcept.io/vertical-spacing/listing-and-listing

### DIFF
--- a/news/239.bugfix
+++ b/news/239.bugfix
@@ -1,0 +1,1 @@
+Remove too large margins on listing blocks @steffenri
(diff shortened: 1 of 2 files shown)


steffenri:
@sneridagh Could you please double-check this? I am a little bit confused when we let the groups-wrapper handle the gaps between blocks, when the block itself handles it and also when to use margin-top and when to use margin-bot? I thought it was always "margin-bot" but it seems that's not always true.

sneridagh:
@steffenri this one was wrong, for sure. There are only special cases (textual blocks mainly) that we allow margin-top, but only on those cases.

BTW, this margin is right, in the case of two appearances of listings, and never on color start. I will take a look this morning.

sneridagh (approved):

PR #259, opened 2023-11-09
Fix css of Sitemap(#209)

### LINKED ISSUE #209: Sitemap is unstyled
It doesn't need a major overhaul, but OOTB the title is positioned to close to the header and too small.

iRohitSingh:
Fix #209 

Screenshot 2023-11-09 at 4 57 49 PM

### DIFF
--- a/src/theme/_sitemap.scss
+++ b/src/theme/_sitemap.scss
@@ -1,9 +1,7 @@
 .section-sitemap {
   #page-sitemap {
     h1 {
-      font-size: $heading1;
-      font-weight: $medium;
-      line-height: 77px;
+      @include page-title();
     }
 
     ul {
--- a/src/theme/_sitemap.scss
+++ b/src/theme/_sitemap.scss
@@ -10,8 +10,8 @@
       li.with-children {
         a {
           font-weight: $bold;
-          text-decoration: none;
           line-height: 45px;
+          text-decoration: none;
         }
         ul {
           li {
--- a/src/theme/_sitemap.scss
+++ b/src/theme/_sitemap.scss
@@ -1,0 +1,26 @@
+.section-sitemap {
+  #page-sitemap {
+    h1 {
+      font-size: $heading1;
+      font-weight: $medium;
+      line-height: 77px;
+    }
+
+    ul {
+      li.with-children {
+        a {
+          font-weight: $bold;
+          text-decoration: none;
+          line-height: 45px;
+        }
+        ul {
+          li {
+            a {
+              font-weight: $regular;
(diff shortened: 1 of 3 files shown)


iRohitSingh:
@sneridagh I have appllied the css for sitemap but it is not center to the container.


Screenshot 2023-11-09 at 5 09 00 PM

sneridagh (changes requested):
@iRohitSingh probably the container it's using is the wrong (Volto's) one. Take a look at

https://github.com/plone/volto/blob/ad9e98633c68d810b0201ba4e9e1c9c8b83e16cf/src/components/theme/View/LinkView.jsx#L25

iRohitSingh:
> @iRohitSingh probably the container it's using is the wrong (Volto's) one. Take a look at
> 
> https://github.com/plone/volto/blob/ad9e98633c68d810b0201ba4e9e1c9c8b83e16cf/src/components/theme/View/LinkView.jsx#L25

@sneridagh regarding this comment even our new a container with default container width applied look like this which is not looking good. can we ask our designer for mockup and then implement it?

Screenshot 2023-11-17 at 5 24 41 PM

src/theme/main.scss
```diff
@@ -31,5 +31,6 @@
 @import 'temp';
 
 @import 'addonsThemeCustomizationsMain';
+@import 'sitemap';
```
ichim-david, line 34: @iRohitSingh @sneridagh  I thought tat addonsThemeCustomizationMain should be the last import in order to get modification from other addons, thus the sitemap import shouldn't be placed here

sneridagh: @ichim-david it is... good catch. @iRohitSingh there can't be anything from that onwards.

sneridagh: @ichim-david fixed. @iRohitSingh please take alook that all is good now.

src/theme/_sitemap.scss
```diff
@@ -0,0 +1,26 @@
+.section-sitemap {
+  #page-sitemap {
+    h1 {
+      font-size: $heading1;
+      font-weight: $medium;
+      line-height: 77px;
```
sneridagh, line 6: Why are we not using the corresponding mixin from the theme for h1?

sneridagh (changes requested):

src/theme/_sitemap.scss
sneridagh: why this need to be redefined?

iRohitSingh: Fixed

PR #261, opened 2023-11-15
Fix header alignment of the sections

sneridagh:
I realised the centering of the menu was not perfect.
I'd like to even improve it a bit more using `react-aria`.


sneridagh:
image

Issue #127, opened 2023-07-04
News Item view


iFlameing:

davisagli:
@iFlameing What was the issue? Is there more to do here?

iFlameing:
@davisagli Now it is fixed.

PR #263, opened 2023-11-18
Added Image Gallery Listing Metadata (#127).

robgietema:


src/icons/fullscreen.svg
ichim-david, line 1: @robgietema @sneridagh  I think that volto-light-theme should introduce a standard for the icon size and keep it consistent across each icon making it easier to size when passing in a width value.

If you look at a dedicated icon library such as remix https://remixicon.com/ each icon has a space of 24x24 to place an icon inside that area.
Pastanaga Icons for instance are all 36x36px.

In this pull request fullscreen.svg and play.svg are 50x50px while left-key right-key are 39x60.

Whichever standard of sizes you want to adopt please keep it consistent and add some info about it

tisto: @ichim-david good point! Let's adopt the 36x36px standard of Pastanaga Icons then.

robgietema: Agree, all icons are now 36x36px.

sneridagh (approved):
LGTM

src/components/Blocks/Listing/ImageGallery.jsx
```diff
@@ -0,0 +1,126 @@
...
+      settings.imageObjects.includes(content['@type']) && content.image_field,
+  );
+  const imagesInfo = renderItems.map((item) => {
+    return {
+      original: `${flattenToAppURL(item['@id'])}/@@images/${
+        item.image_field
+      }/large`,
+      thumbnail: `${flattenToAppURL(item['@id'])}/@@images/${
+        item.image_field
+      }/thumb`,
```
davisagli, line 87: @robgietema @sneridagh We should not be adding new code that uses the shortcut image scale URLs.

PR #268, opened 2023-11-21
Bugfix: Use only one set of values for container widths

ichim-david:
which can be modified from variables


davisagli (commented):
Looks reasonable to me. Any concerns @sneridagh ?

sneridagh (approved):

PR #272, opened 2023-11-22
Remove old code from navigation component. should be a real functiona…

steffenri:
…l component now.


sneridagh (commented):
@steffenri did you copy the code from the updated component in Volto? (I guess so, right?)

PR #276, opened 2023-11-29
Removed the "Overview" Message

Molochem:
As per https://github.com/kitconcept/kitconcept.com/issues/40 the "Overview" React Message is not longer needed


steffenri (commented):
@Molochem 2 things:
1. It's not a bugfix. It's internal. 
2. Run "yarn i18n" to also remove the translation from the .po files.

rest lgtm.

tisto:
Not sure if that is really "internal". Think about someone who uses volto-light-theme. One could even argue that this is a breaking change. Up to the release manager @sneridagh to decide.

davisagli:
Yeah "internal" is for things that are only visible to developers.

sneridagh:
If it's not breaking, then it's a bugfix ;) I think we can go with that one. I didn't like it either, so I'm glad we are getting rid of it.

PR #277, opened 2023-11-29
Fix css of External link icon indicator (#91)

iRohitSingh:
Test-external-icon


sneridagh (approved):
LGTM, @iRohitSingh are we using the last improvements that we have in place in DLR (eg. line drop, etc)? Also, the same policy in show/hide?

PR #280, opened 2023-11-29
Fixes #279 load custom fonts from a separate file and allow setting of cms font instead of hardcoding to Metropolis 

### LINKED ISSUE #279: Custom fonts should be loaded from it's own scss file
If you want to use another set of fonts in another project typography and typo-custom forces you to load Iter and Metropolis whether you want to or not.

ichim-david:


sneridagh (approved):

src/theme/_variables.scss
```diff
@@ -24,8 +24,10 @@ $heading-text-bottom-spacing: 25px !default;
 
 // Fonts
 $custom-fontname: 'Inter' !default;
+$cms-fontname: 'Metropolis' !default;
```
sneridagh, line 27: @ichim-david can they be `cmsui-fontname`? Sorry for the nitty gritty, but I think it's important.

sneridagh (changes requested):
Can we amend slightly the naming of the vars?

PR #282, opened 2023-11-30
Replace static borders with variables

ichim-david:
- Also moved the maps to the top so that any variables after then can make use of the map values such as when I use it in border variables
-  Added defaults also to maps in case someone has other values for them, every variable should end with !default


steffenri:
@ichim-david Can we rename "border-default-XX" to match the rest of the name system? For example we can rename it to "border-solid-small" (then we have the range "border-solid-small", "border-solid-medium" and so on). It' also more descriptive. Although 1px solid might be the default, we should be consistent with our naming.

I am also not so sure about creating variables for specific, single use-cases like you did in "border-table". 

the rest LGTM.

ichim-david:
> @ichim-david Can we rename "border-default-XX" to match the rest of the name system? For example we can rename it to "border-solid-small" (then we have the range "border-solid-small", "border-solid-medium" and so on). It' also more descriptive. Although 1px solid might be the default, we should be consistent with our naming.

@steffenri 
the border-default name is to represent that it is the default border used in the site.
border-solid-small would be an implementation detail for your current design.
Others could choose to use only dotted or dashes styles for their borders.
I welcome another naming suggestion if you would rather use another name to represent the default
border as I think you know as well that naming is the most difficult thing todo especially in a way that
is agreed by everyone.

> 
> I am also not so sure about creating variables for specific, single use-cases like you did in "border-table".
> 
> the rest LGTM.
Normally I would say that if it's a single use case it could be defined in table.scss with default which would allow
the border color to still be different.
I've looked at your figma and the table block has this color as the border color so it's not wild for me to think
that if tables have borders this would be the color to use.

If you don't think that you want to have this border-table variable in variables.scss I could add it in table.scss
with whatever name you want, I welcome a more specific suggestion of how you would want me to change it
so that we can merge the changes.

sneridagh:
@ichim-david @steffenri tbh I had the same feeling about naming, but in a sense that default is the "normal", so it was superfluous. What if we just remove it, then the specialised ones are from the projects?

In the other vars, we are not saying `$spacing-default-small` but `$spacing-small`... So we go with `$border-small`. What do you think?

ichim-david:
@sneridagh ready for another look, I've renamed the border variable to remove default, let me know if this is fine now.

sneridagh (approved):
@ichim-david I made a small adjustment to `xlarge` vars naming, to be consistent with the others. The rest LGTM. 

@steffenri ?

steffenri:
@ichim-david Why do we have 3 different conventions of naming now?
We have
1. border
2. border-{thickness}
3. border-{thickness}-{color}

A border always has a thickness and a color. Why don't our variable names?
I'm all for descriptive variables and "border" can't get any worse.

ichim-david:
> @ichim-david Why do we have 3 different conventions of naming now? We have
> 
>     1. border
> 
>     2. border-{thickness}
> 
>     3. border-{thickness}-{color}
> 
> 
> A border always has a thickness and a color. Why don't our variable names? I'm all for descriptive variables and "border" can't get any worse.

@steffenri  Previously we had border-default but after previous feedback also from @sneridagh where it was mentioned that default is implicit I dropped it.

You have in the site borders
1px, 2px, 4px, 9px and normally black but there are some instances of other colors such as on table or 
most often border: 1px solid black; 

If my site uses 1px 3px 5px 10px solid lightgrey for the borders used in the site because that color works better with my color scheme I have to modify every instance of the static values.
If border-small-black is used everywhere then I will have variable overrides such as
border-small-black: 3px solid lightgrey;
This is why it's good to have a variable that is semantic such as default, table, light, dark, accent ... things that are not color specific and can be modified without feeling weird when you go and read border-small-black: grey;

@steffenri you see the changes, you know the values, give me concrete variable names that you would see using for the site where in variables you can set it to black and yet not hardcode it in the variable name used in the scss files.

src/theme/_variables.scss
```diff
@@ -193,43 +239,20 @@ $grid-block-vertical-spacing-bottom: 80px !default;
...
-  4xl: 48px,
-  5xl: 56px,
-);
+// Border
+$border-width: 1px !default;
+$border-style: solid !default;
+$border-color: $black !default;
+$border-color-light: $greySmoke !default;
+$border-color-table: #c1c1c1 !default;
+$border: $border-width $border-style $border-color !default;
```
danalvrz, line 248: @ichim-david we had a talk about borders, and thought that for now it makes sense to prioritize simplicity. So the only border styles that we want to turn into variables at the moment are the solid black borders. Other borders should be treated as special cases and have their static border values (ex. `border-color` property) set at the level that is needed instead of being global variables. Also, for now maybe let's define `$border` as `1px solid $black`, to avoid the extra width, style and color variables. Thank you for your work! /cc @sneridagh @steffenri

danalvrz (changes requested):

src/theme/_variables.scss
ichim-david: @danalvrz thx for the feedback. I agree with Steffen I don't like $border for the 1px solid black definition.
I think that 1px solid black could have been added to $border-default and had
$border-default
$border-small
$border-medium
and so on but this variable name would be better used as a unit of measure instead of a full border class since
it would need also some meaning such as perhaps a semantic color or section designation.

In order to keep things simple for you guys, I will close this pull request and leave the css as it is.
Having to keep track of what could be deemed as a style value that can have a variable and whatnot is a losing proposition in the long run.

Due to the nature of this theme where the values are the representation of the Figma design file, there is just so much that can be set in a customizable fashion and any extra flexibility would add extra overhead in the maintenance and development of your theme as this pull request has demonstrated when talking about the border solution used throughout the styles.

PR #288, opened 2023-12-05
Upgrade and adapt the theme to`volto-slider-block` 6.0.0 add-on

sneridagh:


tisto:
@sneridagh does that mean we have a hard dependency and the old slider will stop working? I'd be ok with that. However, if that is the case, we should start some kind of compatibility matrix.

sneridagh:
No, it won't stop working. 
The change is in the CSS classes of the main slider elements (not the inner ones). Which is rarely styled, beside margins/paddings.
The good news is that the structure is the same, the classnames are different but they are mapped 1:1 with the old ones. I already documented the mapping in the add-on (see README there).
Only a small adaptation (class bulk rename and CSS check is required. I did the adaptation literally in 15m.
No worries, all is under control.

ichim-david:
@sneridagh I've tested your branch and the slider is indeed a drop-in replacement.
There are a couple of enhancements that I would like but I will add them in the add-on as an issue.

sneridagh:
@davisagli @danalvrz This is ready, please take a final look.

iRohitSingh:
@sneridagh Functionalities by it is working perfectly fine but style is totally missing.


Screenshot 2023-12-15 at 7 39 46 PM


Screenshot 2023-12-15 at 7 38 44 PM


Screenshot 2023-12-15 at 7 36 39 PM


Screenshot 2023-12-15 at 7 35 36 PM


Screenshot 2023-12-15 at 7 46 12 PM

danalvrz (approved):
it works great!

sneridagh:
@iRohitSingh I think you tested the wrong version:

image

Where are the dots coming from?

src/theme/blocks/_slider.scss
```diff
@@ -11,84 +7,33 @@ $sliderImagesAspectRatio: var(--slider-images-aspect-ratio, 16/9);
...
   }
 
-  .teaser-item,
-  .grid-teaser-item {
-    flex-direction: column;
-  }
-  .teaser-item {
-    padding-bottom: $spacing-xlarge;
+  .slider-dots {
+    padding-bottom: 40px;
```
ichim-david, line 15:
```suggestion
    padding-bottom: $spacing-medium;
```

sneridagh:
@ichim-david I reverted your suggestion, because thinking it twice, that padding is fixed, and not dependent of the vertical spacing variable. It's exactly that one for a reason, and it's pixel perfect according with the mocks. It could be 43px, but it's 40px, by chance.

PR #290, opened 2023-12-07
Add Event metadata block

iFlameing:
@sneridagh I tested it out and it is very good. Please look if we need any more things for it.


sneridagh:
@iFlameing I'd like to go for the custom block only for Events, now that we can.
https://gitlab.dlr.de/internet-cms/cms-plone/dlr-internet/-/issues/2276

sneridagh:
@iRohitSingh no semanticui in new components, please! 
You can maintain the `Grid` component only if you want.

iRohitSingh:
@sneridagh I have repeated the css because we have to make decision regarding how we are going to use this block. If you have feedback or how we are going to handle this block.

sneridagh:
@iRohitSingh now (latest Volto) we can do https://github.com/plone/volto/blob/20d6a5efed85c4bdcae66cf4ee594da4b690b14d/docs/source/blocks/settings.md#L211-L217

I already updated it to latest, could you please take care and amend the initial config in the theme? Something like:

```
config.blocks.initialBlocks = {
  Event: [
    { '@type': 'title' },
    { '@type': 'eventMetadata', fixed: true, required: true },
    { '@type': 'slate' },
  ],
};
```

src/components/Theme/EventView.jsx
```diff
@@ -5,22 +5,12 @@
...
-  expandToBackendURL,
-} from '@plone/volto/helpers';
-import {
-  Image,
-  Grid,
-  Button,
-  Container as SemanticContainer,
-} from 'semantic-ui-react';
+import { hasBlocksData, flattenHTMLToAppURL } from '@plone/volto/helpers';
+import { Image, Container as SemanticContainer } from 'semantic-ui-react';
```
iRohitSingh, line 9: @sneridagh Should I remove the Image from semantic one also. I am not removing because we are customizing the volto component.

sneridagh: Please do. Use the main Image component Volto one. Die Semantic!

iRohitSingh: Done.

PR #292, opened 2023-12-12
Fix Image View forces in image content type

iRohitSingh:


iRohitSingh:
Fix #293

news/293.bugfix
```diff
@@ -0,0 +1 @@
+Fix Image View forces in image content type @iRohitSingh
```
davisagli, line 1:
```suggestion
Don't force aspect ratio for the Image content type view @iRohitSingh
```

davisagli (changes requested):

news/293.bugfix
iRohitSingh: ok , thanks

iRohitSingh: Done .

davisagli (approved):
LGTM. @sneridagh @steffenri This is what we always intended for the image view, right?

steffenri:
@davisagli Yes. Only took us like half a year to fix it :D

PR #296, opened 2023-12-12
Update typography patterns to use unitless line-heights

### LINKED ISSUE #273: Light theme should use unitless values as oposed to static absolute values for line-height
https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values

danalvrz:
Fixes #273 

@sneridagh four decimals are the minimum to achieve the correct heights for some designs, but we could always round them if we don't like that. 

I also added a `blockquote()` mixin because I noticed the `blockquote` elements were using SemanticUI styles. As a result, their font size is now bigger, but in fact that is how it is specified on the design spreadsheet, we can always adjust it if needed.

### DIFF
--- a/src/theme/_variables.scss
+++ b/src/theme/_variables.scss
@@ -346,6 +346,7 @@ $line-heights: (
 
 @mixin external-link-icon() {
   display: inline;
-  content: '\a0'url("data:image/svg+xml,%3Csvg aria-hidden='true' focusable='false' data-prefix='fas' data-icon='external-link-alt' class='svg-inline--fa fa-external-link-alt fa-w-16' role='img' xmlns='http://www.w3.org/2000/svg' viewBox='0 0 512 512' width='11' height='11' %3E%3Cpath fill='%23007EB1' d='M432,320H400a16,16,0,0,0-16,16V448H64V128H208a16,16,0,0,0,16-16V80a16,16,0,0,0-16-16H48A48,48,0,0,0,0,112V464a48,48,0,0,0,48,48H400a48,48,0,0,0,48-48V336A16,16,0,0,0,432,320ZM488,0h-128c-21.37,0-32.05,25.91-17,41l35.73,35.73L135,320.37a24,24,0,0,0,0,34L157.67,377a24,24,0,0,0,34,0L435.28,133.32,471,169c15,15,41,4.5,41-17V24A24,24,0,0,0,488,0Z'%3E%3C/path%3E%3C/svg%3E");
+  content: '\a0'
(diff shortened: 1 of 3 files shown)


src/theme/_variables.scss
```diff
@@ -218,15 +218,15 @@ $font-sizes: (
...
-  xs: 16px,
-  s: 18px,
-  m: 24px,
-  l: 30px,
-  xl: 33px,
-  2xl: 36px,
-  3xl: 42px,
-  4xl: 48px,
-  5xl: 56px,
+  2xs: 1.1428,
```
sneridagh, line 221: @danalvrz could you put a comment to hint the final value? :)

sneridagh (changes requested):

src/theme/_variables.scss
sneridagh: btw, if we change the line heights, it's a breaking change.

ichim-david: > @danalvrz could you put a comment to hint the final value? :)

@sneridagh the final value is always relative to the font size used.
Usually, there aren't as many line-height values as there are defined
in this  project, for instance, bootstrap has only 4 values: 1.2, 1.25, 1.5, 2
https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L629

Usually, you have a body text of 1.5, 1.25 for headings, and 1.2 for big headings.

In your design, you have a lot more values from 1 to 1.428
Let's take introduction text as an example of why the value changed from:
xl to 3xl
L - font size 24px
in figma you have:
font-size: 1.5rem;
line-height: 2.0625rem;
16 * 1.5 = 24px
16 * 2.0625 = 33px
33 / 24 = 1.375
this is why @danalvrz changed from xl which was 33px static to 1.375 which
is not.
If I change the l font-size value to 26px, the 1.375 value will keep the line height proportion
which will result in 35.75px.

You can add comments about previous values but in general the line heights are about
proportions and not static values.
I hope I made some sense to you, your request can still be fulfilled if it makes it
easier for your devs to think of what to use.

sneridagh:
I've been looking into it, and see the state of the art and recommendations about the subject. To be honest, I thought that when zooming it would be all borked out, and it is not the case.

https://www.aleksandrhovhannisyan.com/blog/dont-use-a-fixed-line-height/

The thing is that it seems that nowadays browsers adjust the ratio on zoom also, so unit-less values are not that crucial.

The example in

https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

does not apply to us, because we are explicitly setting that per typography tag, not globally.

Our design is very specific about typography, and it's curated to the last intents. The browsers do a good job nowadays with zooming, and the values are set in the typography tag itself, very granular and specific, so this is not a problem to my eyes.

In lieu of simplicity, I'd not touch this and leave the values as it is. If anybody needs to adjust the typography, they will have to override the typography variables anyways, so they will have to adjust the line-heights manually on its own. If they want to use unit-less vars, it's their choice.

sneridagh:
@ichim-david it does not says either that do not use them at all because it's bad :) 

The values are in variables, one can change them in their projects to their liking, so I'd say let's leave it as it is.

@danalvrz please state if some of the naming improvements that you found in your way are worth to keep.

danalvrz:
@sneridagh I think it was a good exploration, but it seems to make sense for things to stay how they were for now. I don't think we should make any naming changes either.

danalvrz:
Closing this PR, since it was decided to keep the line-heights with unit values.

PR #310, opened 2023-12-19
Remove id selector from listings because of css specificity

steffenri:


sneridagh (approved):

sneridagh:
@steffenri it comes from https://github.com/kitconcept/volto-light-theme/pull/263/files#diff-9c12dfa63cebd1052336ef4a22648b3301c2a666f2d43a1fa694ddf8af7107fbR2

@robgietema do you recall why the use of `#main` in here? Could be that removing is going to break something else?

sneridagh:
@steffenri merging, after reading your comment. But why does it break?

steffenri:
@sneridagh Well the fact that we use !important everywhere requires us to overwrite stuff in the projects with higher specificity. When we drastically increase the specificity in VLT (e.g. adding an ID-selector) then we lose control of the overwrites.

PR #311, opened 2023-12-19
Fix showing up uploaded logo from site controlpanel

iFlameing:
@sneridagh If you upload a larger image, the logo will take full width and height of an image. What should be the maximum width and height for the logo image? I am thinking of something 250 and 150. I searched google and it is appropriate.

### DIFF
--- a/news/291.bugfix
+++ b/news/291.bugfix
@@ -1,0 +1,1 @@
+Fix showing up uploaded logo from site controlpanel. @iFlameing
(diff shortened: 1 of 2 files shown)


sneridagh (approved):

sneridagh:
@iFlameing difficult question. In fact, the height and the width has to be set to avoid layout shifting. But being the logo size variable, we would have to provide that information from the uploaded image. I think right now, we are not saving this information in the server. But it would be great to have them set in the rendering. However, one might want to adjust it afterwards using custom values... don't know if that would be suitable to have it in the control panel...

sneridagh (approved):

Issue #304, opened 2023-12-17
Allow choosing between the internet and intranet header


tisto:
We have two different header for VLT:

- default "internet" header (icon left, nav centered, search logo right) https://www.figma.com/file/eYUCbivnNrnjXjBDUizSVX/Volto-Light-Theme?type=design&node-id=994-8440&mode=design
- intranet header (icon left, search input field in the center, navigation below centered) https://www.figma.com/file/eYUCbivnNrnjXjBDUizSVX/Volto-Light-Theme?type=design&node-id=998-8774&mode=design

It should be possible to switch between the two headers via JSON configuration file.

# Internet Header

![header](https://github.com/kitconcept/volto-light-theme/assets/632105/688dfaa3-9dba-443a-a0a5-d7eb03f171e6)

# Intranet Header

![header (1)](https://github.com/kitconcept/volto-light-theme/assets/632105/9fd16648-aef8-44c6-822d-dd8c0b015c74)

tisto:
@sneridagh I'd also like to being able to choose between the internet/intranet header in Volto light theme via a JSON config option. Are you ok with that? Then @iFlameing could work on an implementation.

sneridagh:
Fixed by: https://github.com/kitconcept/volto-light-theme/pull/318

sneridagh:
Merged!

PR #318, opened 2023-12-26
Add options to show intranet flag and intranet header.

iFlameing:
@tisto  @sneridagh screenshot

Screenshot 2023-12-26 at 9 16 25 PM


sneridagh (approved):
LGTM, @tisto ?

davisagli (commented):
This is a nice feature, but it could also be useful to show a label that is not related to an intranet, like showing a banner on a staging site or something.

So, I'd suggest naming the option something more generic, like siteLabel.

README.md
```diff
@@ -181,6 +181,14 @@ It's behind a feature flag, as opt-out:
 config.settings.enableFatMenu = true;
 ```
 
+### Show Intranet Lable
```
davisagli, line 184: The correct spelling is "label" -- please fix it everywhere.

davisagli (changes requested):

tisto:
@davisagli we discussed the naming for the "intranet" header and decided to stick with the "intranet" naming. We were not able to come up with a better name (e.g. one that would describe that the search input field is in the middle). Therefore using the "intranet" flag/label makes sense to me. However, if you have a better idea for the naming, feel free to make a suggestion.

davisagli:
@tisto I don't think we're talking about the same thing; this PR doesn't have anything to do with the search input. It's adding a setting for text to be shown in a label at the top of the site. That text could be labeling it as an intranet, or it could be used for some other purpose.

tisto:
@davisagli there is another feature request to allow switching between the "internet header" and the "intranet header":

https://github.com/kitconcept/volto-light-theme/issues/304

This PR should be in sync with the naming convention that we come up with. It does not make much sense to me to start with this PR and do not have the switc in place. The flag only makes sense to me in combination with the intranet header. No clue why @iFlameing went with that order...

davisagli:
> The flag only makes sense to me in combination with the intranet header. 

@tisto Why? To me, the ability to add the flag sounds like a useful feature whether you are building an intranet or something else (assuming the text can be customized, like it is implemented here).

tisto:
@davisagli tbh I did not think about that option. Therefore we also did not take this into account when designing the header. We would have to amend the top spacing of the "internet header" if we use the flag. The internet header has less vertical space compared to the intranet header. In that case, we would need to design and test four options. We are currently lacking the fourth option as a design specification.

In my personal opinion, we should support two use cases (internet/intranet) with VLT and be opinionated about what we think is the best fit for those use cases. Therefore we should name them internet and intranet and not choose generic names or offer generic options. The flag should only be available on the intranet header option and it does not make much sense to me for the internet option.

Coming up with the "flag" was because of an intranet-specific use-case because we wanted to differentiate between intranet and internet (at FZJ where they use Plone for both).

I might think in a too narrow and specific way here. Though, on the other hand, we have this tendency in the Plone community to make things overly complex by thinking about imaginary use cases. I try to follow the rule of thumb here: if you don't have this use case and nobody asked for it, don't even think about implementing it.

Anyways. I might need to think about this a bit more...

tisto:
@davisagli here are the intranet/internet header. You can see that they have a different height:

Intranet:

intranet-header

Internet:

internet-header

davisagli:
@tisto Yes, I can see how implementing a more generic flag text option would require more design work, and not worth it unless there is an actual use case.

There are some related use case that have come up with DLR:
1. a "staging" banner to make the staging site obvious
2. a customizable banner to show a message about upcoming downtime (we haven't implemented this yet)
But, I think both of those use cases require a full-width banner that would not affect the spacing of the header, so it makes sense to consider them separately from this publicly visible intranet flag.

So I agree with just having a simple boolean setting to switch between 2 header components for intranet and internet -- but that means this PR is going in the wrong direction.

tisto:
@davisagli I agree. The use cases you mention make sense to me, so let's add tickets for those. I agree that this PR might go in the wrong direction. We should implement the intranet/internet header first before tackling the flag use case.

iFlameing:
@tisto @davisagli I will the change the boolean to intranetHeader and if it is there. Then we will have completely different header as well as intranet flag.

iFlameing:
@tisto @sneridagh @davisagli I went through the design of intranet header and implemented it in this pr. 

screenshot

Screenshot 2024-01-09 at 5 36 58 PM

I haven't change the label name. I think it is good. Once it set we know that it is intranet site and we just show the header  of intranet. If you have any feedback or suggestion please mention it.

README.md
iFlameing: done.

iFlameing:
@sneridagh @davisagli If you both review this, I can work on your feedback in coming days.

sneridagh:
@iFlameing please push for this. Question us during the wrapup if needed, but this cannot stay here forever. We are unlikely going to go after developers PRs.

sneridagh (changes requested):
The banner literal should be intl'd. Try to come up with a good DX story for it, please.

iFlameing:
@sneridagh I will push this in this week.

iFlameing:
@sneridagh can you please review this?

cc @davisagli

sneridagh (changes requested):
@iFlameing I set 

```
  config.settings.siteLabel = 'Intra net';
  config.settings.intranetHeader = true;
```




and the label is empty. Maybe some condition is not ok? Also, I'm missing a Cypress test. Do you know that we can "touch" the config from them, right?

iFlameing:
@sneridagh I am adding cypress test which is missing.

iFlameing:
@sneridagh Not able to write a cypress test because cy.settings not working as excepted.

davisagli:
> Not able to write a cypress test because cy.settings not working as excepted.

@iFlameing Give more information if you want help with this. What did you try, and how did it fail?

sneridagh:
@davisagli @iFlameing let's not spend much more time on this. I wanted Alok to try out in case it worked at first shot. Let's look for it in another occasion.

sneridagh (approved):

PR #329, opened 2024-01-18
handle bg color at block level poc

danalvrz:
@sneridagh this is what I was looking into, I checked it against all pages in https://light-theme.kitconcept.io/vertical-spacing and it is very close to rendering exactly the same. There are a few details pending, the mayor one being that plain text doesn't get BG color assigned (only in edit mode, it currently depends on the blocks-group-wrapper BG), maybe we should change that? As said, it is meant as an exploration. Let me know what do you think.


sneridagh:
@danalvrz 

First of all, let me note that tackle this the way we want it is extremely complicated, as shown the last years. And coming up with the best solution is complicated. The combinatory out of it is daunting... and we often overlook it. I wept blood tears while trying to develop it, and the block grouping was the result of not willing to deal with that any longer.

So, the point that you are trying to do is that we might not need grouping, and we could spare the `block-group-wrapper`. If so, then you should remove it in the first place, and make tests without it. Then test in view and edit mode the outcome.

The real challenge here is to write less code to implement what we want, for both edit (and add) and view mode.
The edit mode has its quirks (if you ask me, I would implement it in a far far different way) and it has it's flaws too. But we have to bear with it for now.

The most important question here is: Which block model do you have in mind? We need to adhere to one, that is consistent and the same for all blocks. It can have exceptions, but not every block being one. Think that it's not only you or me touching this CSS or enhance it in projects. The model has to be easy, and understandable, and consistent.

Also, are you aware of all the rules of vertical spacing, right?
In DLR:
https://github.com/kitconcept/dlr-internet/blob/main/docs/development/layout-rules.md

in VLT (subset, but extracted all of them from DLR):
https://github.com/kitconcept/volto-light-theme#vertical-spacing-rules

Also, overlooking margin collapses is wrong, it's in there from the beginning of CSS for a reason. Going all paddings has also its annoyances and edge cases, and it's difficult to teach (since the default is going margins).

Almost all these rules are implemented directly when using a wrapper.

Let's talk about it today!

sneridagh:
@danalvrz https://github.com/kitconcept/volto-light-theme/issues/330

sneridagh:
This has been transferred to the `blockModelv3` branch, closing.

PR #341, opened 2024-02-05
fixed a11y Caption Issues

Molochem:
Fixed the Caption for Images, Video Blocks etc. to use semantically correct HTML Structure

Old Caption
image

New Caption 
image


sneridagh (approved):

news/341.bugfix
davisagli, line 1: @sneridagh Isn't this breaking?

tisto: @sneridagh @sneridagh didn't we agree that changing the HTML structure is a breaking change?

PR #346, opened 2024-02-15
Fix header tabbing order (#251)

### LINKED ISSUE #251: [A11y] header has wrong tabbing order
Tab order on left to right language is expected to be done from left to right.

On https://light-theme.kitconcept.io/
Couple of issues:
- I would expect when tabbing around for first links to go to "Sitemap" and "Github"
followed by the Plone logo.
- I would expect to tab from the menu to the search icon and instead, I tab to the sitemap and github
- I shouldn't be able to tab to 9 since it's not something visible
- There is a jump between Sitemap, Search and Github when tabbing

Here is the tabbing order for the frontpage both logged in and auth, see from Firefox accessibility tab
light-themefull-page-anon
light-themefull-page-auth

iRohitSingh:
Fix #251

### DIFF
--- a/news/251.bugfix
+++ b/news/251.bugfix
@@ -1,0 +1,1 @@
+Fix header tabbing order @iRohitSingh
--- a/src/components/Header/Header.jsx
+++ b/src/components/Header/Header.jsx
@@ -28,18 +28,6 @@ const InternetHeader = ({ pathname, siteLabel, token, siteAction }) => {
   return (
     <>
       <div className="header">
-        <div className="logo-nav-wrapper">
-          <div className="logo">
-            <Logo />
-          </div>
-          <Navigation pathname={pathname} />
-          <MobileNavigation pathname={pathname} />
-          <div className="search-wrapper navigation-desktop">
-            <div className="search">
-              <SearchWidget />
-            </div>
-          </div>
-        </div>
         <div className="tools-wrapper">
           <LanguageSelector />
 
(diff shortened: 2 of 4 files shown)


iRohitSingh:
https://github.com/kitconcept/volto-light-theme/assets/61353484/681f7ccd-401b-4c0a-87e1-3266aa429d5b


https://stackoverflow.com/questions/61065275/firefox-does-not-allow-to-focus-anchor-tag-with-role-button#:~:text=If%20you%20intend%20for%20this,they%20are%20valid%20as%20stated).


There is difference between the tab order of firefox becuase it doesn't support a tag. Please read it here. if you check on chrome it is working perfectly fine.

ichim-david:
@iRohitSingh you still have tabbing that will happen in the 13 tab key because the search-bar isn't hidden and there are elements to tab to although you can't see it because it's with position absolute and with negative margin to hide from view.
Just because it's hidden from view doesn't mean it's hidden from the tab key.
See this screenshot
search-bar-visible-as-13

And see how 13 disappears if I add display: none
search-bar-display-none

This might affect your sliding logic and you will need to add extra code to take this into consideration, for you to check and adjust code as necessary

iRohitSingh:
Welcome-to-volto-light-theme(1) (1)

ichim-david:
@iRohitSingh it looks ok what you did, it would be awesome if a feature branch could be deployed with the content that is usually found at https://light-theme.kitconcept.io/ it would make testing easier for me or someone else by not having todo anything other than checking the live demo instead of spinning up everything locally and creating content to test.

iRohitSingh:
@sneridagh Ready for merge.

sneridagh (approved):

sneridagh:
@iRohitSingh I just found out (by upgrading my site) that this was a breaking change.

So if someone customized the header, then the order of things will change.

We have to be very careful with this changes, the VLT is specially sensible, to our sites and to the other people's sites. Breaking a site on a minor is serious and undermines our credibility. It's easy: are you changing HTML? Then it's breaking. Unless said otherwise, all our public packages should follow:

https://6.docs.plone.org/volto/release-management-notes/index.html?highlight=breaking#definition-of-breaking

Let's revert this, and try to do it right the next time.

PR #366, opened 2024-03-12
fix link in introduction block being smaller than normal text

jonaspiterek:


sneridagh:
@danalvrz could you please take a look?

src/theme/blocks/_introduction.scss
```diff
@@ -19,6 +19,10 @@
     em span {
       font-style: italic;
     }
+    > a {
+      font-size: 24px;
```
tisto, line 23: Just wondering. Can't we somehow incorporate the font-size and line-height from the standard text? Or even better get rid of a fixed setting of those values? Why are links styled differently than regular text in any case?

danalvrz: @jonaspiterek I think moving the `> a` selector up one nesting level should work, so it uses the `introduction()` mixin that is there already. Have you tried this?

sneridagh: @jonaspiterek status?

jonaspiterek: Updated & tested. /cc @danalvrz

PR #371, opened 2024-03-24
set font-weights, font-sizes and font-heights as overridable

erral:
I am not a SCSS expert, but without this I can't modify the settings of the `body-text()` mixin. This way I can override them in my theme's variables.scss and let `body-text()` set a custom font weight or size.

I don't know if this is the right way to do this... but it's the way I found, so disregard it if it makes no sense.


erral:
Or perhaps it would be better to use the variables with weight definitions in the font-weight definition and make the same with the -sizes and line-heights?

danalvrz (approved):
@erral @sneridagh it would be important to make it easy for these maps to be adjusted if needed at a project level, and I think this is a good way. Thank you!

PR #372, opened 2024-03-24
make font definition maps use variables

erral:
See #371 for the original idea

### DIFF
--- a/packages/volto-light-theme/src/theme/_variables.scss
+++ b/packages/volto-light-theme/src/theme/_variables.scss
@@ -79,6 +79,29 @@ $bold: 700 !default;
 $extra-bold: 800 !default;
 $bolder: 900 !default;
 
+// Sizes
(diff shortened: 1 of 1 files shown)


sneridagh:
@erral Regarding this one, we decided back in the day that the main VLT feature are the special curation of the typography and the vertical spacing used. They are quite inter-related and makes no sense to make them easy to change them. Then VLT looses all his reason to be. 

Also, VLT does not aim to be a full blown design system, nor a CSS framework. But this thin layer where the relationships between blocks and text are well defined and curated.

Another thing is to make it easy to override (like your other PR), which we should make easy to do, but I don't think that bloat VLT with variables and make it super configurable is the way to go. We can discuss about it in the next Volto Team meeting.

sneridagh:
@erral I would close this one if you don't mind.

PR #382, opened 2024-04-19
Complete BBB and rename flags

sneridagh:


sneridagh:
@danalvrz I think it's ready to merge with the main feature branch, since we can deal with the Cypress tests at the end.

danalvrz (approved):

PR #384, opened 2024-05-14
Add support for visual regression tests

reekitconcept:
- add cypress plugin dependency
- add cypress configuration files for visual targets
- configure Makefile
- configure github plugin for CI
- customize and fix image snapshot generation
  - customize snapshots folder location
  - fix scrollability of css for full-page screenshots
  - fix lazy image loading
  - disable hover effect on images
  - disable hyphenation (currently disabled but code is available)
  - limit browser to firefox only
- test utilities
  - generate series of tests for different resolutions
  - generate series of tests visiting a list of urls
  - cumulative testing

- add base.js test that
  * tests two pages in anonymous view and serves as an example
  * test a list of pages as an example
  * same in authenticated mode


sneridagh:
@reekitconcept Looks awesome! LGTM, however, I'm missing some docs, afterwards, let's merge it right away.

- how to add new tests/snapshots
- how (if we can) run locally, etc...

@tisto We need to talk to you about how to deal with that last ones.

PR #399, opened 2024-06-25
Update Readme with compatibility info

fredvd:
between Volto and Volto Light Theme versions


fredvd:
I think I forgot to git pull first, rebase becomes a mess. Retry.

PR #430, opened 2024-11-07
Add SemanticUI load under a layer

sneridagh:


packages/volto-light-theme/src/customizations/@root/theme.js
```diff
@@ -1,7 +1,6 @@
 // This customization allows to enable the scape hatch to use
 // SCSS as a theme to override the SemanticUI theme.
-import 'semantic-ui-less/semantic.less';
-import '@plone/volto/../theme/themes/pastanaga/extras/extras.less';
+import './loadSemanticUI.less';
```
davisagli, line 3: @sneridagh uh, this file is missing?

sneridagh: @davisagli it was leaked in the `vlt6` branch, I fixed it now.

packages/volto-light-theme/src/theme/_temp.scss
```diff
@@ -1,2 +1,19 @@
...
+
+  .tab .header h2 {
+    font-size: 16px;
+  }
+}
+
+.block.slider .ui.button.delete-button {
+  padding: 0;
+  border: none;
+}
```
davisagli, line 19: Are these related to the layer?

sneridagh: yes, there are some places that the VLT styling take precedence and override the Pastanaga theme (because they used to be highly specifc, and with a layer, they do not.

davisagli:
@sneridagh There are failures in the acceptance tests. I didn't take a look.

PR #435, opened 2024-11-30
Fix image block schema

danalvrz:


packages/volto-light-theme/news/435.bugfix
```diff
@@ -0,0 +1 @@
+Fix Image block schema @danalvrz
```
davisagli, line 1: @danalvrz Can you be more specific? This PR is hard to review because I don't know what you're trying to fix.

danalvrz: @davisagli I just updated it, thanks! With the new widget and image options on VLT 6 there was an issue with undefined properties when nesting an Image in a Grid block. Whit this change we enhance the Image schema more accurately based on its context.

packages/volto-light-theme/news/435.bugfix
```diff
@@ -0,0 +1 @@
+Add standAloneImageBlockSchemaEnhancer to prevent sidebar setting conflicts between in-Grid and standalone images @danalvrz
```
davisagli, line 1: How about more simply:

```suggestion
Don't show image size and alignment settings for image block inside a grid. @danalvrz 
```

davisagli (commented):
@sneridagh Seems to work ok in my testing. Could you have a look as well?

davisagli (approved):

Issue #438, opened 2024-12-04
External URL icon has strange underline 


Tishasoumya-02:
https://light-theme.kitconcept.io/block/text-block
<img width="582" alt="image" src="https://github.com/user-attachments/assets/ec088fb4-3315-42ed-ba35-b86b26e8b029">

PR #439, opened 2024-12-04
fix: #438 External URL icon has strange underline

Tishasoumya-02:
image

PR #446, opened 2024-12-10
use CSS mask for external link icon

danalvrz:
@Tishasoumya-02 with the intention of being able to use the color system we have in place, maybe using a CSS mask for the icon could work. Please take a look and let me know what do you think.

### DIFF
--- a/packages/volto-light-theme/src/theme/_variables.scss
+++ b/packages/volto-light-theme/src/theme/_variables.scss
@@ -424,8 +424,12 @@ $line-heights: (
 }
 
 @mixin external-link-icon() {
-  $icon: '007EB1';
-  display: inline;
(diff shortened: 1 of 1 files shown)


Tishasoumya-02:
@danalvrz , this logic works a lot better, Thank You for the help . I wasn't aware of this `mask` property . So thanks!

Tishasoumya-02 (approved):

packages/volto-light-theme/src/theme/_variables.scss
```diff
@@ -424,8 +424,12 @@ $line-heights: (
...
+  display: inline-block;
+  width: 11px;
+  height: 11px;
+  background-color: var(--link-foreground-color, #0070a2);
+  content: '';
+  mask: url("data:image/svg+xml,%3Csvg aria-hidden='true' focusable='false' data-prefix='fas' data-icon='external-link-alt' class='svg-inline--fa fa-external-link-alt fa-w-16' role='img' xmlns='http://www.w3.org/2000/svg' viewBox='0 0 512 512'%3E%3Cpath d='M432,320H400a16,16,0,0,0-16,16V448H64V128H208a16,16,0,0,0,16-16V80a16,16,0,0,0-16-16H48A48,48,0,0,0,0,112V464a48,48,0,0,0,48,48H400a48,48,0,0,0,48-48V336A16,16,0,0,0,432,320ZM488,0h-128c-21.37,0-32.05,25.91-17,41l35.73,35.73L135,320.37a24,24,0,0,0,0,34L157.67,377a24,24,0,0,0,34,0L435.28,133.32,471,169c15,15,41,4.5,41-17V24A24,24,0,0,0,488,0Z'/%3E%3C/svg%3E")
```
stevepiercy, line 432: @danalvrz @Tishasoumya-02 I noticed a few differences, and wanted to make sure they were deliberate or if they even matter.

- `viewBox='0 0 512 512'%3E%3Cpath` lost a space between `512'` and `%3E`.
- `fill='%23#{$icon}'` removed.
- `%3E%3C/path%3E%3C/svg%3E");` changed to `/%3E%3C/svg%3E")`.

If these are OK, then I approve. I like that you pulled out some parameters from the difficult to read `url` value and exposed them better.

stevepiercy:
Does this PR supersede #439 and close #438?

packages/volto-light-theme/src/theme/_variables.scss
danalvrz: @stevepiercy the only one that could matter is removing the `fill` parameter, but that is intentional and it will default to `currentColor` if needed.

PR #453, opened 2024-12-19
Fix Slider content is underlined in mobile & Slider button color

### LINKED ISSUE #452: Slider content is underlined in mobile & Slider button color is wrong
Screenshot 2024-12-18 at 8 02 27 PM

iRohitSingh:
Fix #452 

Screenshot 2024-12-19 at 6 01 24 PM

### DIFF
--- a/packages/volto-light-theme/news/452.bugfix
+++ b/packages/volto-light-theme/news/452.bugfix
@@ -1,0 +1,1 @@
+Fix Slider content is underlined in mobile & Slider button color @iRohitSingh
--- a/packages/volto-light-theme/src/theme/blocks/_slider.scss
+++ b/packages/volto-light-theme/src/theme/blocks/_slider.scss
@@ -30,6 +30,11 @@ $sliderImagesAspectRatio: var(--slider-images-aspect-ratio, 16/9);
   }
   .teaser-item {
     padding-bottom: $spacing-medium;
+    a {
+      @media only screen and (max-width: $computer-width) {
+        text-decoration: none;
+      }
+    }
 
     .teaser-item-title {
       position: absolute;
@@ -103,6 +108,7 @@ $sliderImagesAspectRatio: var(--slider-images-aspect-ratio, 16/9);
     border: 1px solid currentColor;
     border-radius: 0;
     background-color: transparent;
+    color: $white;
     font-size: 16px;
     font-weight: 900;
     line-height: 20px;


davisagli:
There is already a rule here which is supposed to prevent underlining links inside a teaser, but it doesn't work in this case because the `a` tag is inside the `.teaser-item` instead of the other way around: https://github.com/kitconcept/volto-light-theme/blob/main/packages/volto-light-theme/src/theme/_bgcolor-blocks-layout.scss#L27

I think we should fix that rule OR fix the markup, instead of adding an override. @sneridagh could you take a look?

iRohitSingh:
> https://github.com/kitconcept/volto-light-theme/blob/main/packages/volto-light-theme/src/theme/_bgcolor-blocks-layout.scss#L27

I also saw this code and get confused, since the markup is different I written the new css for that.

@sneridagh  Here is the code for slider. We are already customised it. If you want I can change the template put the markup other way around. https://github.com/kitconcept/volto-light-theme/blob/main/packages/volto-light-theme/src/components/Blocks/Slider/DefaultBody.jsx#L99C25-L99C37

Also this css is little bit problamatic. If you put anything inside a with not direct children class it will apply the text-decoration and the only options is to override it. But most of time this is what anybody will do. A good solution only put text-decoration when it only contains text node no direct block level element. In this way it guranteed that a only contains presentational element like icon and text.  Or provide a class-name like for this interactive element. Like we do for external icon.

davisagli:
@sneridagh Isn't linking block elements an anti-pattern from an a11y perspective?

https://css-tricks.com/block-links-are-a-pain-and-maybe-just-a-bad-idea/
https://css-tricks.com/block-links-the-search-for-a-perfect-solution/

davisagli:
@sneridagh @iRohitSingh Unfortunately I think it's probably too big of a breaking change to change the markup in volto-slider-block. However, looking again at the rule https://github.com/kitconcept/volto-light-theme/blob/main/packages/volto-light-theme/src/theme/_bgcolor-blocks-layout.scss#L27, I think we could just add the "item" class to this link in volto-slider-block.

sneridagh:
@davisagli I vaguely remember that we did it this way because of something. I can't remember what exactly, maybe DLR asked explicitly for having the link for the whole block rather than for the inner elements?

We are still in alpha, we could still break things. We will have to fix/improve it at some point, specially when this already bite us a couple of times already.

For now, I'll merge the fix for .com, then when we have time, let's decide on this one after Christmas.

iFlameing:
@sneridagh we need to visit this one.

danalvrz:
@iRohitSingh @sneridagh @davisagli It seems like we could update the Slider block to use the new block model, and make the necessary changes for the markup at the same time. Maybe we could even avoid overwriting the body component in VLT with the new version. I think https://github.com/kitconcept/volto-light-theme/pull/453#issuecomment-2557259134 presents a good general solution. I see there is https://github.com/kitconcept/volto-slider-block/pull/50, could it be continued? In the meantime, the original [issue](https://github.com/kitconcept/volto-light-theme/issues/452 ) was fixed at css level in https://github.com/kitconcept/volto-light-theme/pull/509/commits/d8384dd04acd4b76b8a9b60705c4192dc0a96fc6#diff-ec18d4872da61a811ff8fb1c684bf8a64ba16b007fd35653ebfd36bf56405d1b.

sneridagh:
+1 to fix it in slider.

PR #454, opened 2024-12-22
Color contrast check

danalvrz:
This PR adds a color contrast check to the theming widget to help with accessibility. If the color contrast ratio between background and foreground colors is < 4.5 it displays a warning message under the color picker with the current color contrast ratio, the WCAG compliance level & a link to further info about color contrast accessibility.

Screenshot:
 
Screenshot 2024-12-22 at 9 59 55 a m


iFlameing:
@danalvrz @sneridagh shouldn't instead of putting everything on plone site. I simple controlpanel provided regarding the theming and assets. So that everything is at one place. Also, we can also make give them some config directly in that controlpanel.  Giving them proper cms feels instead of touching code. Just my thoughts after seeing this.

sneridagh:
@danalvrz let's try to merge this soon, could you please solve conflicts and merge?
Also, could you imagine to add some Cypress tests to it?

sneridagh:
@danalvrz let's talk later, but I've switched for now to a new color picker widget, we might have to transfer all the code there or make it "shareable" between widgets (my preferred option).

davisagli:
Replaced by #463

PR #462, opened 2025-01-30
Listing variations with summary component

davisagli:
What's included here:
- Re-label the built-in listing variations to List (was Default), List with images (was Summary), and Grid.
- Add a new Summary component which is looked up by content type and renders the kicker, title, and description
- CSS tweaks to make the kicker look correct where it was not shown before
- Add Summary component implementations for News Item, Event, and File
- Adjust date rendering in EventView, NewsItemView, and EventMetadata block for better consistency
- Docs


packages/volto-light-theme/src/components/Summary/DefaultSummary.jsx
```diff
@@ -0,0 +1,16 @@
+const DefaultSummary = (props) => {
+  const { item, HeadingTag = 'h3' } = props;
+  return (
+    <>
+      {item?.head_title && <div className="headline">{item.head_title}</div>}
+      <HeadingTag className="title">
+        {item.title ? item.title : item.id}
+      </HeadingTag>
```
davisagli, line 8: This keeps the same tag that was used before (h2 or h3) for backwards compatibility, but ideally we would not use those tags, since this isn't part of the page outline (table of contents).

packages/volto-light-theme/src/components/Summary/DefaultSummary.jsx
```diff
@@ -0,0 +1,16 @@
+const DefaultSummary = (props) => {
+  const { item, HeadingTag = 'h3' } = props;
+  return (
+    <>
+      {item?.head_title && <div className="headline">{item.head_title}</div>}
```
davisagli, line 5: I really wish we could switch this class to use `kicker` instead of `headline`, especially since `headline` is also used for the heading at the top of the listing block.

packages/volto-light-theme/src/components/Summary/NewsItemSummary.jsx
```diff
@@ -0,0 +1,34 @@
...
+        {item.title ? item.title : item.id}
+      </HeadingTag>
+      {!item.hide_description && (
+        <p className="description">{item.description}</p>
+      )}
+    </>
+  );
+};
+
+export default NewsItemSummary;
```
davisagli, line 34: There's quite a bit repeated between the default summary and the custom ones for News Item and Event -- when really we are only changing the kicker. Should we be adding a Kicker component instead?

On the other hand, doing it this way makes it possible to show something entirely different than the title + description if that makes more sense for some content type.

packages/volto-light-theme/src/components/Blocks/Listing/DefaultTemplate.jsx
```diff
@@ -24,16 +23,22 @@ const DefaultTemplate = ({ items, linkTitle, linkHref, isEditMode }) => {
...
-              </div>
-            </ConditionalLink>
-          </div>
-        ))}
+        {items.map((item) => {
+          const Summary =
+            config.getComponent({
+              name: 'Summary',
+              dependencies: [item['@type']],
+            }).component || DefaultSummary;
```
davisagli, line 31: Does using a component in the registry just make it harder to explain than updating an object in config.settings (i.e. like contentTypeViews)?

packages/volto-light-theme/src/components/Summary/NewsItemSummary.jsx
```diff
@@ -0,0 +1,34 @@
+const NewsItemSummary = (props) => {
+  const { item, HeadingTag = 'h3' } = props;
+  const formatter = new Intl.DateTimeFormat(item.Language || 'default', {
+    year: 'numeric',
+    month: 'long',
+    day: 'numeric',
+  });
```
davisagli, line 7: We ought to update the FormattedDate component in volto to use this instead of momentjs, but that'll have to be in a major version...

packages/volto-light-theme/src/helpers/dates.js
```diff
@@ -0,0 +1,24 @@
+import { parseISO } from 'date-fns';
```
sneridagh, line 1: @davisagli Why do we need `date-fns`? 
In core, the `FormattedDate` component does not need it at all.
It costed a big deal to remove all the useless date libraries from Volto, (we had `date-fns` in there), I won't introduce it again if there's a good enough vanilla JS counterpart.

docs/summary.md
```diff
@@ -0,0 +1,119 @@
...
+
+const BlogPostSummary = (props) => {
+  const { item, HeadingTag = 'h3' } = props;
+
+  const effective = parseDateFromCatalog(item.effective);
+  const headline = [
+    effective ? (
+      <FormattedDate
+        key="day"
+        date={effective}
```
sneridagh, line 56: As far as I remember, `FormattedDate` is able to format ISO dates.

docs/summary.md
```diff
@@ -0,0 +1,119 @@
...
+        <p className="description">{item.description}</p>
+      )}
+    </>
+  );
+};
+
+export default BlogPostSummary;
+```
+
+The custom summary component must be registered in the Volto registry in your add-on's `applyConfig` function.
```
sneridagh, line 87: One of the things that I discussed with @stevepiercy several times is to standardize glossary and concepts. We haven't settled yet on one, but I'd call this "default configuration loader" and try to be consistent everywhere.

packages/volto-light-theme/src/components/Blocks/EventMetadata/View.jsx
```diff
@@ -1,20 +1,30 @@
...
   const content = props.properties;
-  const dateOptions = {
-    year: 'numeric',
-    month: 'long',
-    day: 'numeric',
-    hour: 'numeric',
-    minute: 'numeric',
-  };
-  const intl = useIntl();
+  const start = content.start ? parseISO(content.start) : null;
```
sneridagh, line 11: Why do you need to parse it first?
I checked and `FormattedDate` does:
```
const toDate = (d) => (typeof d === 'string' ? new Date(d) : d);
```

is it not enough?

packages/volto-light-theme/src/components/Blocks/Listing/DefaultTemplate.jsx
sneridagh: IMHO, I think we should API-fy all `@plone/registry`, stop people pointing to create entries directly in the config. eg. I'd create `registerWidget`, `registerBlock`... and so on.

packages/volto-light-theme/src/components/Summary/DefaultSummary.jsx
sneridagh: +1

packages/volto-light-theme/src/components/Summary/NewsItemSummary.jsx
sneridagh: @davisagli I'd do it. I think that in general, we should slice every time a piece of structure/content has to be reusable (even if it's for one time).

packages/volto-light-theme/src/helpers/dates.js
davisagli: @sneridagh I am using it to solve a specific problem.

If I parse an ISO date without the time, then vanilla JS assumes it is UTC, and we end up rendering the wrong date here in GMT-7:
```
>>> value = new Date("2025-02-10")
Date Sun Feb 09 2025 16:00:00 GMT-0800 (Pacific Standard Time)
>>> new Intl.DateTimeFormat("en", {year: 'numeric', month: 'long', day: 'numeric'}).format(value)
"February 9, 2025"
```

Ah, but it looks like I can make it work without `date-fns` if I specify UTC as the timeZone when I format:
```
>>> value = new Date("2025-02-10")
Date Sun Feb 09 2025 16:00:00 GMT-0800 (Pacific Standard Time)
>>> new Intl.DateTimeFormat("en", {year: 'numeric', month: 'long', day: 'numeric', timeZone: 'UTC'}).format(value)
"February 10, 2025"
```

So let me see if I can update it to work that way. (But it means we need to keep track of whether we're dealing with a datetime or just a date.)

Core Volto has the same problem btw.

docs/summary.md
davisagli: @sneridagh I'm happy to update it once there's a glossary entry I can link to.

packages/volto-light-theme/src/components/Blocks/Listing/DefaultTemplate.jsx
davisagli: @sneridagh Doesn't that make it harder to do things like replacing a core block by extending its config?

packages/volto-light-theme/src/components/Summary/DefaultSummary.jsx
davisagli: I added a note about this to the listing block issue for Seven.

packages/volto-light-theme/src/components/Summary/NewsItemSummary.jsx
davisagli: @sneridagh I'm not sure I follow; where else would the Kicker be used aside from in the summary component? We can't use the same Kicker component on the content view, because the data is structured differently in the summary serialization. I'm going to leave it as is for this PR.

packages/volto-light-theme/src/helpers/dates.js
davisagli: @sneridagh You're right, we don't need `date-fns`. I updated the PR. Things work okay in most cases since we're getting an ISO date including the tz offset from the backend. There's a separate issue in plone.restapi related to including the tz offset for the effective date which I need to come back to.

packages/volto-light-theme/src/components/Blocks/Listing/DefaultTemplate.jsx
sneridagh: @davisagli could be... but then we will have a `updateBlock`? This is something that @ericof and I talked several times, and we felt like a good idea.

PR #475, opened 2025-02-12
Be more defensive against imperfect blocks in RenderBlocks

teekuningas:
With empty gridBlock and the new themes we get this when rendering blocks:
```
Cannot read properties of undefined (reading 'themes')
```
I guess it's because "empty" type has not been taken into account. But as a easy fix, perhaps add a "?" to not crash.


sneridagh (approved):

sneridagh:
@teekuningas thanks for contributing!

PR #489, opened 2025-03-11
set parent block theme as nesteded blocks theme

danalvrz:


sneridagh:
@danalvrz has to be paired with: https://github.com/plone/volto/pull/6832

sneridagh (approved):

PR #491, opened 2025-03-13
Updated objectListWidget using JSONField as an array

sneridagh:
----
📚 Documentation preview 📚: https://volto-light-theme--491.org.readthedocs.build/

### DIFF
--- a/docs/reference/helpers.md
+++ b/docs/reference/helpers.md
@@ -15,3 +15,56 @@ This chapter lists the helper utilities in Volto Light Theme.
 
 The successor of Volto's `DragDropList`, using `dnd-kit` library.
 
+Given a `items` array of objects, and a children render prop function, it allows you to reorder them using drag and drop.
+
+It has the following props:
+
+```ts
+interface DndSortableListProps {
+  /**
+   * An object containing all items, ideally with a `@id` key with the item's
+   * unique identifier (UID).
+   *
+   * If none provided, the item index in the array is used.
+   *
+   */
+  items: Array<Record<string, any>>;
+
+  /**
+   * A function to handle the end of a drag operation.
+   * @param {DragEndEvent} event - The drag end event.
+   */
+  handleDragEnd: (event: DragEndEvent) => void;
+
+  /**
+   * The currently active object being dragged.
+   */
+  activeObject: any;
+
+  /**
+   * A function to set the active object.
+   * @param {any} object - The object to set as active.
+   */
+  setActiveObject: (object: any) => void;
+
+  /**
+   * A render prop function to render each sortable item.
+   * @param {object} params - The parameters for rendering the item.
+   * @param {string} params.uid - The unique identifier of the item.
+   * @param {any} params.item - The item data.
+   * @param {number} params.index - The index of the item in the sorted list.
+   * @param {any} params.activeObject - The currently active object being dragged.
+   * @param {object} params.attributes - The attributes for the sortable item.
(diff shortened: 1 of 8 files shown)


docs/reference/helpers.md
```diff
@@ -15,3 +15,56 @@ This chapter lists the helper utilities in Volto Light Theme.
 
 The successor of Volto's `DragDropList`, using `dnd-kit` library.
```
stevepiercy, line 16:
```suggestion
`DndSortableList is the successor of Volto's `DragDropList`, using the `dnd-kit` library.
```

docs/reference/helpers.md
```diff
@@ -15,3 +15,56 @@ This chapter lists the helper utilities in Volto Light Theme.
...
 
+Given a `items` array of objects, and a children render prop function, it allows you to reorder them using drag and drop.
```
stevepiercy, line 18:
```suggestion
Given an array of objects as `items`, and a `children` render prop function, you can drag and drop the items to reorder them.
```

stevepiercy (changes requested):
Just docs review.

packages/volto-light-theme/src/helpers/DndSortableList.tsx
```diff
@@ -105,6 +105,11 @@ const DndSortableList = (props: DndSortableListProps) => {
     setActiveObject(null);
   }
 
+  const sortedItems = items
+    ? items.map((item, index) => item['@id'] || index)
+    : [];
+  const list = items ? arrayToObject(items) : [];
```
davisagli, line 111: Can we call it itemsByUid instead of list? Calling something `list` when it's an `object` is very confusing to my Python-focused brain. Also it should default to `{}` rather than `[]` if there are no items. Doesn't TypeScript catch that sort of thing?

packages/volto-light-theme/src/helpers/DndSortableList.tsx
```diff
@@ -105,6 +105,11 @@ const DndSortableList = (props: DndSortableListProps) => {
     setActiveObject(null);
   }
 
+  const sortedItems = items
```
davisagli, line 108: Sorted is the wrong name for this, because there's no sort(). It looks like it's an array of the item ids in order, so maybe call it `order` or `itemIds`?

packages/volto-light-theme/src/helpers/DndSortableList.tsx
sneridagh: `arrayToObject` was not fully typed correctly, and the fact that is returning any[] didn't help.

I fixed it.

packages/volto-light-theme/src/helpers/DndSortableList.tsx
sneridagh: Fixed.

sneridagh:
@davisagli FYI, I tried what we talked about removing the `@id` on save, but I could not make it work. Not even with your workaround, I guess the fact that you did it "outside" the widget, then the own widget lifecycle does not get in the way. Problem that happened: every time you typed something, it looses the focus in the widget field...

If you want to take a look, go ahead.

sneridagh:
@davisagli merging for now too.

davisagli:
@sneridagh I saw about the `@id`. It could also be a breaking change to remove it, so it's okay to keep it as long as it's not causing problems.

sneridagh:
@davisagli yeah, I thought that also, this way it's a drop-in replacement, and works everywhere.

PR #492, opened 2025-03-14
Add actions to header + Fix hydration problem in /edit + Sitemap container

sneridagh:
@ericof up to you to merge after the repo refactor.


sneridagh:
@ericof starting to merge things, but the release steps are not running, apparently, for no reason. They run during the monorepo PR... but not after your last PR (take a look at the logs in the monorepo PR). Now I don't know if this is intended or not.

Merging, just to know if they run in main... which I can't see that in the workflow definitions :(

PR #494, opened 2025-03-17
Update to use `@inherit` endpoint

sneridagh:
Paired with https://github.com/kitconcept/kitconcept.intranet/pull/77 and merge with https://github.com/kitconcept/volto-light-theme/pull/492


packages/volto-light-theme/src/components/Footer/FooterLogos.tsx
```diff
@@ -73,7 +71,7 @@ const FooterLogos = () => {
               if (!logoInfo.src) return null;
 
               return (
-                <li className="item" key={logoInfo.href}>
+                <li className="item" key={logoInfo['@id']}>
```
davisagli, line 74: `@id` isn't set in logoInfo. Should this be src?

davisagli:
See the eslint errors

packages/volto-light-theme/src/components/Footer/FooterLogos.tsx
sneridagh: yes, it's the item `@id` from the widget.

davisagli: @sneridagh logoInfo is initialized above with empty values. It does not include `@id`. I think you wanted to get the `@id` from `logo`.

PR #498, opened 2025-03-19
Update highlight block

sneridagh:
----
📚 Documentation preview 📚: https://volto-light-theme--498.org.readthedocs.build/


sneridagh:
@ericof when the changelog is only docs, should we follow the same approach like in Volto? Just put it somewhere?

PR #519, opened 2025-04-01
Update to Plone 6.1.1

davisagli:
Let's see if my updates to the github workflow are correct...I'm trying to use the Makefile as the source of truth for which Plone version to use.


ericof:
@davisagli Please add a changelog fragment on the backend folder

davisagli:
@ericof To be clear: I care about having a single place to update the Plone version, if possible. I don't care whether that place is the Makefile or not. It could be a version.txt like we have in kitconcept.intranet, but I didn't see that until after I worked on this.

PR #523, opened 2025-04-04
Added sticky_menu

sneridagh:


frontend/packages/volto-light-theme/src/components/Widgets/schema/stickyMenuSchema.ts
```diff
@@ -0,0 +1,98 @@
...
+        title: 'Default',
+        fields: ['icon', 'alt', 'href', 'openLinkInNewTab'],
+      },
+    ],
+
+    properties: {
+      title: {
+        title: intl.formatMessage(messages.title),
+      },
+      icon: {
```
davisagli, line 74: @sneridagh Note, `icon` will not be found automatically by the resolveuid transform or linkintegrity.

sneridagh: @davisagli As long as the URL is in the field `@id` (objectBrowser) is fine:

From the serializer (blocks.py)
```python
    def _process_data(self, data, field=None):
        if isinstance(data, str) and field in self.fields:
            return uid_to_url(data)
        if isinstance(data, list):
            return [self._process_data(data=value, field=field) for value in data]
        if isinstance(data, dict):
            fields = ["value"] if data.get("@type") == "URL" else []
            fields.append("@id")
            fields.extend(self.fields)
            newdata = {}
            for field in fields:
```

and the result:

```
[
 {
  "@id": "3e939bea-ce96-42b6-83b5-d15441b22041",
  "icon": [
   {
    "@id": "resolveuid/eec82559bf3242a6be4d43bc2096f399",
    "@type": "Image",
    "Description": " The Image content type can be used to upload an image in various formats (JPG, GIF, PNG, SVG). The uploaded image should always have a high resolution so that it can be used flexibly, for example as a banner image. Plone automatically delivers the images in the best scaling, so there is no need to scale images down manually.",
    "Title": "Image - Light",
    "image_field": "image",
    "title": "Image - Light"
   }
  ],
  "alt": "asd",
  "href": [
   {
    "@id": "resolveuid/c40a2be214b645338a7ef2e711e2f5c6",
    "Description": "",
    "Title": "Vertical Spacing",
    "title": "Vertical Spacing"
   }
  ]
 }
]
```
So I think we are good!!

sneridagh:
Waiting for some design blockers to resolve.

backend/news/523.feature
```diff
@@ -0,0 +1 @@
+Added behavior `voltolighttheme.sticky_menu`. @sneridagh
```
davisagli, line 1: It is actually named kitconcept.sticky_menu

PR #526, opened 2025-04-22
Minor typo fix in message

danalvrz:


davisagli (commented):
@danalvrz Don't we need a translation for this?

PR #536, opened 2025-05-07
Fixed props for Summary component in Teaser view

sneridagh:


frontend/packages/volto-light-theme/src/components/Blocks/Teaser/DefaultBody.jsx
```diff
@@ -29,7 +29,7 @@ const TeaserDefaultTemplate = (props) => {
   const Summary =
     config.getComponent({
       name: 'Summary',
-      dependencies: [data['@type']],
+      dependencies: [href['@type']],
```
davisagli, line 32: It looks like `href` can be undefined if there isn't an item selected yet. line 24 should be careful and fallback to an empty object.

davisagli (changes requested):

PR #538, opened 2025-05-07
Add actions to card

danalvrz:
This is a test to have actions (a button, or badges/tags) in the Card component. Also, it considers the case in which the Card does not have a `target` link.


danalvrz:
closing in favor of #539

PR #540, opened 2025-05-12
Added some of Volto's acceptance tests

sneridagh:
@iRohitSingh Could you please take a look and make sure that all passes?
Could be that we have to amend some of them, to match VLT things.
Also, it would be great to include Volto's slate tests too, just in case.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -132,6 +132,8 @@ jobs:
         include:
           - name: "Main"
             short-name: main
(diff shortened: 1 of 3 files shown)


iRohitSingh:
Only the Cypress test for the search block is left. I’ll take care of it.

iRohitSingh:
@sneridagh  I have fixed all the cypress problems, but merging main broke some more test. I will work on those today.

PR #547, opened 2025-05-15
Fix footer backgroundcolor

sneridagh:
@danalvrz we have a conditional background colors now, depending if the main footer is present or not, and it does not apply to the whole footer. Did you add it because of something?


danalvrz:
I am missing background color in the Footer the way it is, but maybe the color goes somewhere else?:
Screenshot 2025-05-15 at 10 40 50 a m

PR #567, opened 2025-06-05
Fixed summaries on Events and News Items

sneridagh:


frontend/packages/volto-light-theme/src/components/Blocks/Teaser/schema.js
```diff
@@ -1,5 +1,8 @@
 export const teaserSchemaEnhancer = ({ schema, formData, intl }) => {
   schema.properties.href.selectedItemAttrs.push('getRemoteUrl');
+  schema.properties.href.selectedItemAttrs.push('effective');
+  schema.properties.href.selectedItemAttrs.push('start');
+  schema.properties.href.selectedItemAttrs.push('end');
```
davisagli, line 5: @sneridagh I think it's a design flaw that these need to be configured (instead of including all metadata that came from the server). Once the block is saved, if it's a live teaser then the metadata is updated in the teaser block serializer, which includes the full summary serialization (https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/blocks.py#L250)

PR #581, opened 2025-06-13
Reintroduce an improved Anontools using portal_actions.

sneridagh:
@davisagli I opted for not to enable the login (login is not enabled by default, it seems in Plone). We could think if we would like to (I guess not).

Should we backport this to 6?

### DIFF
--- a/docs/reference/slots.md
+++ b/docs/reference/slots.md
@@ -37,7 +37,7 @@ The `FooterLogos` component will be rendered in the `preFooter` slot of the page
 
 `headerTools`
 :   The `headerTools` slot adds content in the header tools at the top-most right area of the page.
-    By default, it has the `Anontools` component, which provides the login, logout and register links if they are enabled in the [`portal_actions`](https://6.docs.plone.org/backend/portal-actions.html) tool in your Plone site.
+    By default, it has the `Anontools` component, which provides the login, logout, and register links if they are enabled in the [`portal_actions`](https://6.docs.plone.org/backend/portal-actions.html) tool in your Plone site.
 
 `preFooter`
 :   The `preFooter` slot adds content before the footer of the page.
--- a/frontend/packages/volto-light-theme/news/581.feature
+++ b/frontend/packages/volto-light-theme/news/581.feature
@@ -1,1 +1,1 @@
-Reintroduce an improved `Annontools` using `portal_actions`. @sneridagh
+Reintroduce an improved `Anontools` using `portal_actions`. @sneridagh
--- a/news/581.documentation
+++ b/news/581.documentation
@@ -1,1 +1,1 @@
-Reintroduce an improved Annontools using portal_actions. @sneridagh
+Reintroduce an improved `Anontools` using `portal_actions`. @sneridagh
--- a/backend/news/581.feature
+++ b/backend/news/581.feature
@@ -1,1 +1,1 @@
-Reintroduce an improved Annontools using portal_actions. @sneridagh
+Reintroduce an improved `Anontools` using `portal_actions`. @sneridagh


docs/conceptual-guides/site-customization.md
```diff
@@ -144,6 +144,8 @@ If you have enabled the intranet header, the intranet flag is the text in the gr
 You can define the actions located at the top right of the header.
 These are links to other pages, each of which is defined by the fields for a title, a target URL, and a boolean for whether or not to open the link in a new tab.
 
+The special actions `login`, `logout`, and `register` are shown if they are enabled in your Plone site as actions.
+These can be defined in the `Actions` control panel in the Plone site or programmatically via Generic Setup.
```
stevepiercy, line 148:
```suggestion
These can be defined in the {guilabel}`Actions` control panel in the Plone site or programmatically via Generic Setup.
```

docs/reference/slots.md
```diff
@@ -37,6 +37,7 @@ The `FooterLogos` component will be rendered in the `preFooter` slot of the page
 
 `headerTools`
 :   The `headerTools` slot adds content in the header tools at the top-most right area of the page.
+    By default, it has the `Anontools` component, which provides the login, logout and register links, if enabled as actions in your Plone site.
```
stevepiercy, line 40:
```suggestion
    When enabled as actions in your Plone site, by default it has the `Anontools` component, which provides the login, logout, and register links.
```

frontend/packages/volto-light-theme/news/581.feature
```diff
@@ -0,0 +1 @@
+Reintroduce an improved Annontools using portal_actions. @sneridagh
```
stevepiercy, line 1: OK, but what does "improved" mean? Please be specific so developers know what was actually changed.

```suggestion
Reintroduce an improved `Annontools` using `portal_actions`. @sneridagh
```

stevepiercy (changes requested):

docs/reference/slots.md
sneridagh: It has the `Anontools` component set. This component has the ability to show these links depending on the portal_actions defined in that moment.

sneridagh: I find the suggestion quite disturbing, I don't know what are we talking about after reading it.

davisagli: @stevepiercy 
```suggestion
    By default, it has the `Anontools` component, which provides the login, logout and register links if they are enabled in the portal_actions tool in your Plone site.
```

stevepiercy: Where's "the `portal_actions` tool in your Plone site."? I don't see it.

https://demo.plone.org/controlpanel/

davisagli: @stevepiercy portal_actions is a tool in the ZMI from ancient times. It's typically configured by a developer using actions.xml in a GenericSetup profile. (Well, usually you don't have to change the default actions.) There is also an Actions control panel in Classic UI, but it hasn't been added to the REST API or Volto yet.

davisagli:
> I opted for not to enable the login (login is not enabled by default, it seems in Plone).

@sneridagh What do you mean? It is enabled here: https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/profiles/default/actions.xml#L435

> Should we backport this to 6?

I'd say so, since we also accidentally removed the login link there when we added the header_actions

docs/reference/slots.md
stevepiercy: I think this is something that only experienced Plone developers know about. I don't know if you want to apply the tenet from Seven to VLT docs, specifically, "write documentation as if you don't have years of experience with Plone".

Fortunately, I found three possible resources for the term `portal_actions` to make that job easier.

https://6.docs.plone.org/search.html?q=portal_actions

Each of the three is interesting because each has its unique context. I'm not sure whether any are relevant here. Perhaps a `seealso` directive with a link to relevant docs would be helpful?

What do y'all think?

sneridagh:
> @sneridagh What do you mean? It is enabled here: https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/profiles/default/actions.xml#L435

Not for Volto sites, this is demo.plone.org:

image

I searched for it and I could not find where we remove it.

news/581.documentation
```diff
@@ -0,0 +1 @@
+Reintroduce an improved Annontools using portal_actions. @sneridagh
```
stevepiercy, line 1:
```suggestion
Reintroduce an improved `Anontools` using `portal_actions`. @sneridagh
```

sneridagh:
@stevepiercy applied!

PR #592, opened 2025-07-03
Improve color handling for highlight & introduction blocks

danalvrz:


danalvrz:
@sneridagh We need https://github.com/kitconcept/volto-highlight-block/pull/14 and a new release to fix the failing test. Since we are not forcing the white font at VLT level anymore for the variation 2 of the Highlight block, it is not passing the contrast check.

danalvrz:
@sneridagh I would consider this only affects internal implementation, not external behavior.

PR #593, opened 2025-07-04
First listfromlist widget implementation

sneridagh:


davisagli (commented):
For a11y we need some way to move and order items that doesn't depend on drag and drop. (It also doesn't look very obvious to me that you can do it with drag and drop.)

PR #601, opened 2025-07-09
Move all listing template to use Card component

iFlameing:
I will write changelog and all the ci test once everything is ready.


iFlameing:
> @iFlameing it is looking very good, I only found a couple issues when trying it locally. (1) If I select the "List with images" variation, the links on the items disappear so I can't navigate anywhere when clicking on an item. (2) If I select the "Event calendar" variation on the example content page, it gives me an error.

Fixed the navigation issue for the list with images ones.
2. Event calendar has never been used for listing, so I disabled the selection for now. If we need we have to amend the template to support the lisitng ones also.

iFlameing:
@sneridagh you can now review this.

frontend/packages/volto-light-theme/src/primitives/Card/Card.tsx
```diff
@@ -93,18 +93,22 @@ type CardImageProps = {
   /** A custom React component to render the image. */
   imageComponent?: React.ComponentType<any>;
   children?: React.ReactNode;
+  showDefaultImage?: boolean;
```
danalvrz, line 96: I would probably name this maybe "showPlaceholderImage" to me it clearer.

danalvrz (commented):
looks good, I only added a comment about naming.

frontend/packages/volto-light-theme/src/components/Summary/EventSummary.tsx
```diff
@@ -6,7 +6,7 @@ import FormattedDate from '@plone/volto/components/theme/FormattedDate/Formatted
 import type { DefaultSummaryProps } from './DefaultSummary';
 
 const EventSummary = (props: DefaultSummaryProps) => {
-  const { item, HeadingTag = 'h3', hide_description } = props;
+  const { item, HeadingTag = 'h3', a11yLabelId, hide_description } = props;
```
sneridagh, line 9: We got along without having to do this in the teasers:

https://github.com/kitconcept/volto-light-theme/blob/f89239135e191812d7c738cab3859e18763253b0/frontend/packages/volto-light-theme/src/primitives/Card/Card.tsx#L80

It has to work without it. Please remove it from al Summaries.

frontend/packages/volto-light-theme/src/primitives/Card/Card.tsx
sneridagh: +1

frontend/packages/volto-light-theme/src/config/blocks.tsx
```diff
@@ -302,7 +302,10 @@ export default function install(config: ConfigType) {
 
   config.blocks.blocksConfig.gridBlock.blocksConfig.listing.variations =
     cloneDeep(
-      listingBlockVariations.filter((variation) => variation.id !== 'grid'),
+      listingBlockVariations.filter(
+        (variation) =>
+          variation.id !== 'grid' && variation.id !== 'eventCalendar',
```
sneridagh, line 307: Is there any `eventCalendar` variation? why?

sneridagh (changes requested):
Please take a look at the comments.

Next will be after merging: Move the acceptance tests to kitconcept.intranet.
Make sure that all the example content follows there too. Make sure that the a11y tests work too. The usual... :)

frontend/packages/volto-light-theme/src/config/blocks.tsx
iFlameing: we are using listingbody component which fetch the listing template accessing the variation. so yes we have eventCalendar variation so that listing body can picked up.

frontend/packages/volto-light-theme/src/primitives/Card/Card.tsx
iFlameing: done

PR #618, opened 2025-07-24
Fix width of listing body in grid listing variante

iFlameing:
Added one line .card flex-grow:1, and refactor the inner and summary inside it to make css at one place.


iFlameing:
@sneridagh PR for morning bug we have seen in timo q/a.

PR #622, opened 2025-07-28
Fix: Logos Block Size Issue

Tishasoumya-02:


frontend/packages/volto-light-theme/src/theme/_footer.scss
```diff
@@ -355,17 +355,38 @@ ul.logos-container {
   grid-template-columns: repeat(auto-fit, minmax(300px, 1fr));
   list-style: none;
 
+  .logo-image {
+    display: block;
+    width: 100%;
+    height: auto;
+    object-fit: contain;
```
iFlameing, line 362: why do you have written object-fit: contain. Because it get overwritten by img classname.

iFlameing (commented):
Everything seems to fine. can you please remove this and then it is good.

iFlameing:
Also please create example-content and update the branch with main.

PR #625, opened 2025-08-12
Add RSS block styles

danalvrz:


frontend/packages/volto-light-theme/news/+addRssBlockStyles.feature
```diff
@@ -0,0 +1 @@
+Add styles for RSS block. @danalvrz
```
davisagli, line 1: This is a bit misleading, it's really only styles for the specific RSS block variations included in kitconcept.intranet. They won't work if someone just installs VLT and volto-rss-block and doesn't customize the variations.

davisagli (approved):

PR #626, opened 2025-08-21
Fix Teaser center aligned extra padding-bottom

Tishasoumya-02:
https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-intranet/-/issues/117

image


sneridagh (commented):
@Tishasoumya-02 I do not understand, there are two teasers in a row (in the screenshot). So it cannot have that classnames

Tishasoumya-02:
@sneridagh @iFlameing I think I understood why this is happening, it's because after the teaser a separator block is added which increases the extra space , if we remove it, then it's back to normal spacing



https://github.com/user-attachments/assets/d9bac4d1-2fc3-4193-9800-403d639750e5

sneridagh:
@Tishasoumya-02 ah, I see, the separator should not be needed, right? if two teasers are in a row, the bottom bar is applied by the block rules. Can you assess that? If so, this fix is not needed.

/cc @danalvrz

danalvrz:
@sneridagh @Tishasoumya-02 A bottom border is added when there are two contiguous teaser blocks, this is a custom behavior (similar to how two contiguous grid blocks change their spacing). If I look at the [example page](https://light-theme.kitconcept.io/block/teaser-block), the layout matches our designs. So I will close the PR.

PR #638, opened 2025-09-01
Use image widget in `preview_image_link` fields

sneridagh:
@Tishasoumya-02 can you take over?
The idea is to use the `image` widget for the `preview_image_link`.
I guess that it's left to do amend tests, and create new ones (that cover the direct upload from the widget).
Also I saw that the CSS is a bit off on hover over the cross SVG when trying to remove the image from the widget... could you take a look too? Thanks!

image

image


sneridagh:
@Tishasoumya-02 You are correct, it's not present in VLT... but in intranet (kitconcept/core has it by default in all content types)
Let's merge this, and then amend the intranet. In the meanwhile you can work on it in the intranet if you use mrs-developer to checkout this branch of VLT. If you have any question regarding that, please ping @iFlameing

sneridagh:
@Tishasoumya-02 I've been taking a look, turns out we had a shadow on the edit component for the image block. The fix in Volto 18.25.0 for the ImageWidget includes a change in there, so we had to update the shadow. That's why if we can save from creating a shadow is always better and something to check always that something weird or unexpected happens.

davisagli:
@sneridagh We need to check more explicitly when we upgrade Volto here to see if any shadowed components need to be updated.

PR #645, opened 2025-09-08
Fix person description text overflow in grid listing and teaser blocks

iRohitSingh:
Grid Listing: 

Screenshot 2025-09-08 at 8 12 23 PM

Grid Teaser: 

Screenshot 2025-09-08 at 8 11 51 PM

### DIFF
--- a/frontend/packages/volto-light-theme/src/theme/person.scss
+++ b/frontend/packages/volto-light-theme/src/theme/person.scss
@@ -297,7 +297,7 @@ body.person-squared-images,
     .summary-extra-info {
       min-width: 0;
       gap: 6px;
-      a {
+      .summary-email {
         display: block;
         overflow: hidden;
         text-overflow: ellipsis;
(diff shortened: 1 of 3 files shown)


frontend/packages/volto-light-theme/src/theme/person.scss
```diff
@@ -290,3 +290,25 @@ body.person-squared-images,
...
   }
 }
+// person with Four columns
+.ui.stackable.stretched.four.column.grid {
+  .card-summary {
+    display: grid;
+    .summary-extra-info {
+      min-width: 0;
+      gap: 6px;
+      a {
```
davisagli, line 300: @iRohitSingh Please add a summary-email class and only add the ellipsis for the email.

(It's not okay to do this for all links inside a 4-column grid, because then there is no way for the user to read the link text. It's okay for the email, because the link text is the same as the link href, so they can get the email by clicking the link.)

davisagli (changes requested):

sneridagh (changes requested):
@iRohitSingh Sorry, I am not ok with the position of the classname you've chosen... Wouldn't be better if you wrap the whole items (as in icon + link) in the classname, so you can address them properly (as a whole, and as a specific elements), if needed in the future?

sneridagh:
@iRohitSingh also, please fix the violation.

PR #650, opened 2025-09-10
FlagAlign->Align and add German translation for the same in Slider Block

Tishasoumya-02:
Ticket : https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-intranet/-/issues/147

### DIFF
--- a/backend/news/+addTranslationsforSliderblock.internal
+++ b/backend/news/+addTranslationsforSliderblock.internal
@@ -1,1 +1,0 @@
-In slider-block replace FlagAlign with Align and add German translation @Tishasoumya-02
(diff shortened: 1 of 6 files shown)


frontend/packages/volto-light-theme/locales/de/LC_MESSAGES/volto.po
```diff
@@ -212,10 +212,10 @@ msgstr ""
 msgid "Field"
 msgstr ""
 
-#. Default: "Flag align"
+#. Default: "Align"
 #: components/Blocks/Slider/schema
 msgid "Flag align"
```
tisto, line 217: @sneridagh as long as we are in alpha with VLT, we can as well change the msgid here, correct? Then I would go for it. Otherwise, we have this inconsistency and will forget about it. What do you think?

davisagli: @tisto @sneridagh I'd say we should create a VLT 8 milestone, add an issue there listing the msgids that we want to break in order to match the current text, and we can follow up on this when there is less time pressure.

stevepiercy: FYI, I suggested a similar strategy for Volto 19 in https://github.com/plone/volto/issues/6159#issuecomment-2450831033. I'd be interested in contributing to this effort. I know words! 😁

PR #651, opened 2025-09-10
Add german translation of slider variant "simple".

iRohitSingh:


frontend/packages/volto-light-theme/news/651.internal
```diff
@@ -0,0 +1 @@
+Add german translation of slider variant "simple" @iRohitSingh
```
davisagli, line 1:
```suggestion
Add German translation of slider variant "simple" @iRohitSingh
```

And please note, translations should be `.bugfix` instead of `.internal` since they are a user-facing change. `.internal` is for things that are only seen by the developers.

davisagli (approved):

PR #658, opened 2025-09-14
Revert prepublish

sneridagh:
@ericof turns out that 

```
  "plonePrePublish": {
    "publish": false
  },
```

Do not publish at all. We need to document these things throughly, in this case, in publishing things with `repoplone` and their relationship with Cookieplone setup and `prepublish` script. So I guess in core, under Cookieplone docs. I plan to work in these in the next weeks.

Also, internal docs, that's on me too.


ericof:
You are correct. Lets create an issue on repoplone to document all the scenarios and what do we want to happen.

PR #666, opened 2025-09-22
Don't show parent tags when adding the child.

iFlameing:


iFlameing:
@sneridagh It is also ready for merge.

PR #674, opened 2025-09-25
Fix cypress test for calendar block.

iFlameing:


frontend/cypress/tests/main/blocks-calendar.cy.js
```diff
@@ -409,16 +432,9 @@ describe('Event Calendar Block Tests', () => {
...
-      .should('have.class', 'has-end-date');
-
-    cy.get('.card-listing')
-      .contains('Second Event')
-      .closest('.card-listing')
-      .find('.image-wrapper .date-inset')
+    cy.get('a[href="/my-second-event"]')
+      .next('div.card-inner')
+      .find('.date-inset')
       .should('not.have.class', 'has-end-date');
```
davisagli, line 438: Now the test doesn't do what the description says. It says "Event calendar show single date for same day event and double for different" but it's not making any assertion to confirm the "and double for different" part of that.

iFlameing: @davisagli I re-added the assertion for date are different. I thought that is unnecessary that's why I remove it.But i was wrong.

PR #678, opened 2025-09-29
Add default value to footer colophon text & style fixes

danalvrz:


sneridagh (changes requested):
@danalvrz I don't know why it's not working, I think the condition is wrong.
@davisagli your proposal didn't include the default value being set?

davisagli:
> @danalvrz I don't know why it's not working, I think the condition is wrong.

@sneridagh Can you make a complete report, what steps you did and what you expected? OR @danalvrz Can you provide instructions on how to test it?

> @davisagli your proposal didn't include the default value being set?

@sneridagh This component has a fallback where it renders "Powered by Plone and Volto Light Theme" along with a copyright and logo. If we add a default value in the backend, then the copyright and logo will never be shown.

sneridagh:
@davisagli @danalvrz 

I start this PR, and the footer is gone:

image

I would expect that the hardcoded default is used.

Then, when I edit it still shows until I set something in the colophon. At this point, it shows what I set.
If I remove the colophon, it shows again the default.

davisagli (commented):
I tested and it's working like I expected:

1. with no VLT backend (I started the backend using `make backend-docker-start` from the volto repository): it shows "Powered by Plone and Volto Light Theme" along with the copyright and logo.

2. with VLT backend and empty "Footer colophon text" -> it shows nothing

3. with VLT backend and a value for "Footer colophon text" -> it shows this text

The one further change we could make is that if we want to still show "Powered by Plone and Volto Light Theme" by default when the VLT backend is installed, we can add it as the default here: https://github.com/kitconcept/volto-light-theme/blob/main/backend/src/kitconcept/voltolighttheme/behaviors/customizations.py#L258 ..this should only take effect for newly created sites.

sneridagh:
@davisagli @danalvrz getting back to this, I would do what David suggested, and add a default for newly created sites, so with a fresh VLT installation, it does show something. However, this would be a breaking change: in the conference site, it won't show once they upgrade the site to VLT8. It should be noted properly in the docs.

davisagli (approved):

sneridagh (approved):

PR #686, opened 2025-10-06
Several fixes Oct BRH

sneridagh:
WIP


davisagli:
I think you need to add the | separator with a pseudo-selector. Otherwise we can only hide the file-size and file-type but not the separators.

PR #702, opened 2025-10-30
NotFound Page add css

Tishasoumya-02:
image


iFlameing (commented):
You missed the font-size of p tag. it is 24px instead of 18px.

sneridagh (approved):

PR #713, opened 2025-11-05
Use volto-dsgvo banner series 3 in VLT 7

sneridagh:
@jackahl Do you see any problem with it? I don't see anything breaking in the Changelog.


sneridagh:
@Tishasoumya-02 any problem that we bump it to series 3?

PR #716, opened 2025-11-05
Native namespaces

sneridagh:


davisagli (approved):

davisagli:
Note: this shouldn't be necessary with the new version of horse-with-no-namespace that I released today. But it should also be no problem to do it.

sneridagh:
@davisagli merging the alpha one at least.

PR #719, opened 2025-11-05
Native namespaces in the backend packages

sneridagh:


davisagli:
Note: this shouldn't be necessary with the new version of horse-with-no-namespace that I released today.

PR #730, opened 2025-11-17
Add german translation for Event Calendar block in block chooser.

iFlameing:
screenshot

blockchooser


davisagli:
Shouldn't these go in https://github.com/kitconcept/volto-calendar-block instead?

iFlameing:
PR in in volto-calendar-block https://github.com/kitconcept/volto-calendar-block/pull/4

PR #752, opened 2025-12-30
fix bug where markdown inline links in descriptions are rendered as a…

jackahl:
… separate paragraph

### DIFF
--- a/frontend/packages/volto-light-theme/src/theme/blocks/_listing.scss
+++ b/frontend/packages/volto-light-theme/src/theme/blocks/_listing.scss
@@ -51,7 +51,7 @@
       object-fit: cover;
     }
 
-    p.description > a {
+    .card-inner a {
       display: inline;
     }
   }


davisagli (commented):
What existing rule sets the links to NOT be inline? Can we make that one more specific instead of adding this override? (There will probably be other links inside listing items in the future which are not in the description.)

sneridagh:
@davisagli default Volto is setting:

```.listing-item, .listing-item a { display: flex }```

Which we cannot modify, since it would be breaking. Regarding specificity, you are correct, how about setting it to `card-inner a` instead, so we cover all the use cases inside the card?

image

davisagli:
> Which we cannot modify, since it would be breaking.

@sneridagh Well, technically we could fix it in Volto 19. I don't think this was ever intended to select more than the main `a` tag for the listing item. But I trust your judgment if you think changing it there would be too disruptive.

> Regarding specificity, you are correct, how about setting it to card-inner a instead, so we cover all the use cases inside the card?

Sure, that sounds like an improvement.

PR #753, opened 2026-01-06
fix the Login-Logout issue - shows login link even when the user is not logged in

Tishasoumya-02:
Issue-https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-intranet/-/issues/212

https://github.com/user-attachments/assets/192fd113-86f3-4e13-a4fd-1d71656873df

We remove ___ac_ cookie after the _auth_token_ cookie has expired and is directed to logout when there is some action on the page. 
For this particular example - I have set the expiry time to a min , the token expires after a minute and when there is some action on the page , it redirects to logout and cookies are deleted as seen in the inspect.

For Volto - The login-logout issue does not occur. I tested it with reducing the token expiry time and on token expiration it logs out and shows the login .

https://github.com/user-attachments/assets/2f530406-0db0-42eb-b9ad-ada2258fe80d

### DIFF
--- a/frontend/packages/volto-light-theme/news/+fixdoublenavincards.bugfix
+++ b/frontend/packages/volto-light-theme/news/+fixdoublenavincards.bugfix
@@ -1,0 +1,1 @@
+Fixed double navigation in cards that contains inner links in its body. @sneridagh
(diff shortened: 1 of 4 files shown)


iFlameing:
> @iFlameing uh, ok... true, this was "fixed" long ago, so it gets consistent with the classic login too.
> 
> @iFlameing @Tishasoumya-02 Given this fact, I am hesitating now if we should fix this in Volto, instead of only in VLT. What do you think?

It make more sense fixing in Volto to make sure we remove backend token when frontend token expire.

davisagli:
Yes, let's please do this fix in Volto. It doesn't affect the AnonTools links there, but it can cause other similar problems if we don't fix it.

PR #785, opened 2026-02-06
Show only two letters of the native language name

iRohitSingh:
Screenshot 2026-02-06 at 8 13 24 PM

### DIFF
--- a/frontend/packages/volto-light-theme/news/785.feature
+++ b/frontend/packages/volto-light-theme/news/785.feature
@@ -1,0 +1,1 @@
+Show only two letters of the native language name @iRohitSingh


frontend/packages/volto-light-theme/src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -0,0 +1,91 @@
...
+            className={cx({ selected: toReactIntlLang(lang) === currentLang })}
+            to={translation ? flattenToAppURL(translation['@id']) : `/${lang}`}
+            title={langmap[langKey]?.nativeName || langKey}
+            onClick={() => {
+              onClickAction();
+            }}
+            key={`language-selector-${lang}`}
+          >
+            {/* START CUSTOMIZATION */}
+            {langmap[langKey]?.nativeName.slice(0, 2) || langKey}
```
davisagli: This solution won't work well in general, because the [2-letter ISO language code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes) is not always the same as the first two letters of the language's full name.

Instead, use the first 2 letters of the langKey

davisagli:
@sneridagh I gave some feedback that had no response yet

PR #807, opened 2026-02-25
fixed the image size being narrow-container-width in edit mode

TimoBroeskamp:
Fixed the size of a single image in a grid block.


sneridagh (approved):
LGTM, I agree that we should track why the image has a default value set in there. But this fix suffices for me.

PR #818, opened 2026-03-11
Improve Card component link handling and style rafactoring

danalvrz:
https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-intranet/-/issues/200


frontend/packages/volto-light-theme/src/components/Summary/DefaultSummary.tsx
```diff
@@ -7,19 +8,43 @@ export type DefaultSummaryProps = {
   HeadingTag?: React.ElementType;
   a11yLabelId?: string;
   hide_description?: boolean;
+  cardHref?: string;
+  cardItem?: Partial<ObjectBrowserItem>;
+  cardOpenLinkInNewTab?: boolean;
+  cardIsInteractive?: boolean;
+  cardPrimaryLinkRef?: React.RefObject<HTMLAnchorElement | null>;
```
davisagli: @danalvrz @sneridagh Doing it this way is awfully complicated for anyone who is creating custom Summary components for a custom content type...they have to understand and add all these props, and add the ConditionalLink which passes them all through correctly.

I'd like to advocate for the LinkToItem pattern I used in https://github.com/kitconcept/volto-light-theme/pull/688 -- that way Summary components just need a small update to receive the LinkToItem and wrap it around the title. The LinkToItem encapsulates a ConditionalLink that is constructed within the top-level Card component and passed down to the Summary (but can still receive children when it is used inside the Summary). I can explain tomorrow if it's not clear.

We should keep the other things you did differently than me, of course.

By the way, we need to update the docs for the Card component.

frontend/packages/volto-light-theme/news/+ImproveLinkHandlingInCard.feature
davisagli, line 1: This is a breaking change, so should be .breaking (that helps put it in the correct section of the changelog).

It's breaking because custom Summary components now need to accept and use the LinkToItem prop, and because the markup changed. We need a new section in the upgrade guide to explain what you might need to update in your project using VLT because of these changes.

We should also explicitly mention the change to `ul` instead of heading tags here in the release note.

frontend/cypress/tests/a11y/gridBlock.cy.js
davisagli, line 1: Why did you remove the a11y test for the grid block?

frontend/packages/volto-light-theme/src/components/Blocks/Listing/SummaryTemplate.jsx
```diff
@@ -51,13 +51,16 @@ const SummaryTemplate = ({ items, linkTitle, linkHref, isEditMode }) => {
                   showPlaceholderImage={true}
                   imageComponent={PreviewImageComponent}
                 />
-                <Card.Summary a11yLabelId={props.a11yLabelId}>
-                  <Summary item={item} HeadingTag="h3" />
+                <Card.Summary
+                  a11yLabelId={props.a11yLabelId}
+                  LinkToItem={props.LinkToItem}
```
davisagli, line 56: The Card automatically passes LinkToItem to all children, so it shouldn't be necessary to do that explicitly here.

davisagli (changes requested):

frontend/packages/volto-light-theme/src/components/Blocks/Listing/SummaryTemplate.jsx
davisagli: Oh right, it's because we're adding ItemBodyTemplate here and have to pass it down.

frontend/packages/volto-light-theme/news/+ImproveLinkHandlingInCard.feature
danalvrz: I changed the extension to breaking, and added the upgrade guides.

frontend/cypress/tests/a11y/gridBlock.cy.js
danalvrz: I thought it was a bit of a duplicate since it is looking at a simple page with a listing, but I have added it again.

davisagli: @danalvrz Oh I see, this is just the listing of pages that actually show the grid, which have their own a11y tests. Yes, it's fine to remove this.

davisagli (commented):
This looks good to me, aside from the failing CI.

Check the things I did differently in https://github.com/kitconcept/kitconcept.intranet/pull/358 to see if any of that seems like something we should copy here.

I'd like @sneridagh to take a look at this before we merge too.

sneridagh (approved):
LGTM!! Awesome work!

davisagli (approved):

PR #821, opened 2026-03-12
Fixed the font-color in the tablehead when using h2 or h3

### LINKED ISSUE #820: Title block font-color changes in tablehead when making it to a Title or subtitle
When using a table and changing the font-size of the tablehead to title or subtitle the font-color changes to white.
Standart: 

Image

With title font-size selected:

Image

with subtitle font-size selected:

Image

Steps to reproduce:
Go here: https://demo.plone.org/block/table-block
Go into Editmode
Select the titlehead and change its fontsize
Click anywhere else so you stop highlighting it

Expected behaviour: For the font-color to stay black when changing to title or subtitle.

TimoBroeskamp:
fixes #820


danalvrz (approved):

TimoBroeskamp:
I found another weird interaction while testing.
Depending on the order of adding a link or making it a heading or subheading the styles are not applied properly.
The following was made to a heading and then I also applied a link to it:
image

The next one was made a link and afterwards made a heading:
image
As you can see in the second image the link gets removed.

danalvrz:
@TimoBroeskamp I was able to reproduce this as well, and the same thing happens in text blocks. So let's open a ticket in gitlab and work on it as a separate issue.

PR #830, opened 2026-03-31
hide sticky menu under the mobile navigation

jackahl:


jackahl:
@danalvrz I decreased the z-index for the sticky menu significantly now. Do you know whether there are any objects, that would potentially conflict with this and now overlap?

danalvrz (approved):

Issue #834, opened 2026-04-09
fat menu navigation closes when clicking on scrollbar


jackahl:
In some cases the navigation fat menu is too tall on desktop to fit the screen height and therefore scrolling is necessary to click at the bottom items.
Some users are using the window scroll bar on the right to scroll, but clicking on it will close the navigation, making it impossible for them to reach the bottom items in the fat nav.

jackahl:
I think this could be fixable in https://github.com/kitconcept/volto-light-theme/blob/5035b58d7df022f649fb62ef504d94c1db0b1a5f/frontend/packages/volto-light-theme/src/components/Navigation/Navigation.tsx#L101

jackahl:
Some reference on how others are handling this: https://github.com/mui/material-ui/issues/15724

TimoBroeskamp:
PR: https://github.com/kitconcept/volto-light-theme/pull/835

PR #835, opened 2026-04-09
Fixed the navigation closing when using the scrollbar

TimoBroeskamp:
Fixed a bug where the navigation closes when using the scrollbar to scroll down.
Issue: https://github.com/kitconcept/volto-light-theme/issues/834


jackahl (changes requested):
@TimoBroeskamp I tested it locally and with this change, the menu does not close at all when just clicking somewhere on the page.

davisagli (approved):

PR #836, opened 2026-04-17
backport to 7.x.x for issue#834

TimoBroeskamp:
This is the backport change to this PR: https://github.com/kitconcept/volto-light-theme/pull/835
Issue: https://github.com/kitconcept/volto-light-theme/issues/834

### DIFF
--- a/frontend/packages/volto-light-theme/news/+fixNavigationClosing.bugfix
+++ b/frontend/packages/volto-light-theme/news/+fixNavigationClosing.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug that closed the Navigation when scrolling down with the scrollbar. @ TimoBroeskamp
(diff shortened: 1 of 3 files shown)


jackahl:
@TimoBroeskamp @sneridagh can we get this merged and released till Thursday?

jackahl (dismissed):
Functionality works, but @TimoBroeskamp please remove this gigantic update of the pnpm-lock.yaml

davisagli (approved):
I cleaned up pnpm-lock.yaml since @TimoBroeskamp is not in today. (I ran `git checkout 7.x.x frontend/pnpm-lock.yaml to reset just that file to the copy from the 7.x.x branch)

This looks ok to me.

davisagli:
@jackahl How did you test? The fix is not working for me in Firefox or Chrome. I opened the Blocks navigation, scrolled a bit with my touchpad to make the scrollbar appear, then clicked it. The fat menu closed, but my understanding was that it should not.

jackahl:
@davisagli I switched on the permanent in the OS settings. It only works with this and not with the temporary scrollbar that appears when you use gesture scrolling or mousewheel. 
But I would argue that people who use the scrollbar by dragging it, will have this setting enabled. And the initial report of the problem was coming from Windows anyways, where there is always a scrollbar as well.

davisagli:
@jackahl ah ok, that makes sense. Let's go ahead with it then.

PR #837, opened 2026-04-17
Fix navigation text color not loading on hard reload.

iFlameing:
I already written down my investigation in ticket see here. https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-website/-/issues/30#note_32280

It seems like a easy fix but it is not. VRT will take care of test but still some manual q/a will be good.

cc @jnptk @danalvrz


jnptk (commented):
LGTM! Tested locally & works fine for me

sneridagh (approved):

PR #838, opened 2026-04-21
Fix Footer aligned with content

iRohitSingh:
Issues Link: https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-website/-/issues/31

Screenshot 2026-04-21 at 2 33 58 PM


sneridagh (approved):

sneridagh:
@iRohitSingh For every fix in VLT UI like this one that does not trigger a VRT (visual regression test) failure means that we don't have one for this, and we should. Please create one accordingly. Merging this one now since it's a sensitive one.

PR #844, opened 2026-04-24
fixed some combinations with the richtext-widget in the text-block

TimoBroeskamp:
Fixed some combinations with the richtext-widget in the text-block.
Issue link: https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-website/-/issues/34

image

image


TimoBroeskamp:
I ran into VR-tests that were failing and the part that is failing is the table of contents block. 
I checked the figma design and currently it is not correctly implemented. [Figma Design](https://www.figma.com/design/eYUCbivnNrnjXjBDUizSVX/Volto-Light-Theme?node-id=2099-25074&t=w2u5uJJ41Ma3UUCJ-0)
In the design it has a line height of 30px and currently its 24px so I will include it in this PR too.
Also I ran into this error because I inherit the line-height and that changed the previous version.

PR #845, opened 2026-04-24
fix subscript and superscript combined with a link

TimoBroeskamp:
Fixed the combination of subscript and superscript combined with a link not being displayed correctly.
Issue link: https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-website/-/issues/37

image


TimoBroeskamp:
What is missing right now is the part of the issue that you cannot create lists in the introduction block. Maybe the fix is to exclude the widget-option to create a list since its also not in the figma design?
Figma design: https://www.figma.com/design/eYUCbivnNrnjXjBDUizSVX/Volto-Light-Theme?node-id=385-5814&p=f&t=w2u5uJJ41Ma3UUCJ-0

PR #846, opened 2026-04-27
Fix the extra padding header-wrapper adding to search-wrapper and center the search

Tishasoumya-02:
Smaller viewport :

image


Tishasoumya-02:
@iFlameing could you take a look here,if it is the right approach to go with. Since the search is inside the <code>.header-wrapper</code> the fix is little restrictive .

Tishasoumya-02:
Translation fix:https://github.com/plone/volto/pull/8153

frontend/packages/volto-light-theme/src/theme/_layout.scss
```diff
@@ -90,6 +90,17 @@ body {
   // This is the pusher for the header under 1440px so over that is tight to the border
   padding: 0 1em;
 
+  //Adjusting for search in header in smaller viewports
+  @media only screen and (max-width: $large-monitor-width) {
```
iFlameing: I don't think this is a good solution. The solution is pretty simple.

We are getting padding from header-wrapper 1em both side so we have to manage it somehow.

The search-bar active is position absolute so I will increase it width from 100% to width: calc(100% + 2em);

then I will make left to -1em. When it is happening when screen size hit the 1440px. so we are going to apply this when max-width is 1440px.

Image

Tishasoumya-02: got it yes , it should work for smaller viewports

frontend/packages/volto-light-theme/src/theme/_header.scss
```diff
@@ -158,6 +158,10 @@
           height 600ms ease-in-out;
         transition-property: top, height;
 
+        @media only screen and (max-width: $large-monitor-width) {
```
iFlameing: also add +2em there to inside max-width to make it perfect.

iFlameing (approved):
LGTM!

jnptk (approved):
LGTM!

Issue #456, opened 2025-01-02
Provide a more traditional menu option for medium sized websites


tisto:
We currently have two options for the fat menu:

- fat menu that opens on click: well suited for large websites, consistent behavior across devices (mobile and desktop)
- no menu at all (only suitable for very small websites)

There are two possible other options we might want to consider:

- fat menu that opens on hover (like on fz-juelich.de). The downside is that mobile and desktop behave differently and that the UX is not very good
- traditional menu that only expands when there is content

tkimnguyen:
I don't know how you'd implement this in Volto but on the Plone side I can imagine a catalog index that tells you how many contained items something has, and if that number is 0 then the click takes you to it, otherwise it shows the fat menu and the contained items. One could argue either way that this is or isn't "more" consistent UI behaviour...

tisto:
If I recall correctly, @fredvd and/or @davisagli looked into creating an index for this. I guess we should make sure there is a PLIP for this.

davisagli:
There was some discussion of an index like this (actually it would be a metadata column) for a different use case (showing a folder icon in the contents view) at the Beethoven Sprint a couple years ago, but I don't remember if anyone started implementing it. The trick is figuring out how to update the index when the number of contained items changes from 0 to 1 or 1 to 0, but avoid unnecessary index updates otherwise.

tisto:
@davisagli right. Thanks for the update. I guess we should write a PLIP then so that we do not forget about this. This metadata column would indeed be useful for both use cases. Can you imagine writing a PLIP for this?

davisagli:
@tisto The previous discussion of catalog indexes was off topic. We already have the information we need in the navigation data. I'll hide those comments so we can keep this issue focused on the UX.

tisto:
@ericof I outlined four options above that are backed by UX research. I still cannot wrap my head around the fifth option (opens on click only if there are subitems; leads directly to the page if there are no subitems). If you want us to discuss this further, I need your help explaining it in more detail to me (in person).

PR #849, opened 2026-05-01
Fat menu: skip opening for top-level items with no children

### LINKED ISSUE #848: Fat menu: Disable opening the menu if no children in that area/section
If the menu item has no children, we should take the user directly to the destination, not open the "empty" menu.

ericof:
## Summary

Fat menu used to open an empty submenu when a top-level navigation item had no children. With this change, those items render as a direct link (`NavItem`) and clicking them navigates straight to the destination.

## Approach

The check uses the existing nav tree — `item.items?.length > 0` — so no catalog metadata or backend changes are required. Per @davisagli's note on #848: for the fat menu we already have the nav tree items, and what matters is "children in the nav structure, not all children".

## Changes

- `Navigation.tsx`: gate the fat-menu UI behind `hasFatMenu && hasItems`; fall through to `NavItem` otherwise.
- `nav.cy.js`: acceptance test creating a top-level page with no children and clicking it from the desktop menu.
- News fragment.

## Test plan

- [x] `make lint` and `make format` clean
- [x] Unit tests pass (`vitest`, 44/44)
- [ ] Cypress acceptance test passes (`pnpm cypress:run` / CI)
- [x] Manual: with fat menu enabled, top-level item with children opens submenu (existing behavior)
- [x] Manual: with fat menu enabled, top-level item without children navigates directly

Closes #848

### DIFF
--- a/frontend/cypress/tests/main/nav.cy.js
+++ b/frontend/cypress/tests/main/nav.cy.js
@@ -61,5 +61,6 @@ context('Navigation Acceptance Tests', () => {
   it('Open destination page if no subitems', function () {
     cy.wait('@content');
     cy.get('ul.desktop-menu a').contains('Just a page').click();
+    cy.get('.documentFirstHeading').should('have.text', 'Just a page');
   });
 });
--- a/frontend/cypress/tests/main/nav.cy.js
+++ b/frontend/cypress/tests/main/nav.cy.js
@@ -26,6 +26,12 @@ context('Navigation Acceptance Tests', () => {
       contentTitle: 'Level 3',
       path: '/level-1/level-2',
     });
+    cy.createContent({
+      contentType: 'Document',
+      contentId: 'a-page',
+      contentTitle: 'Just a page',
+      path: '/',
+    });
 
     cy.visit('/');
     cy.viewport('macbook-16');
@@ -50,5 +56,10 @@ context('Navigation Acceptance Tests', () => {
     cy.get('ul.desktop-menu button').contains('Level 1').click();
     cy.get('.subsubitem-wrapper').findByText('Level 3').click();
     cy.get('.documentFirstHeading').should('have.text', 'Level 3');
+  });
+
+  it('Open destination page if no subitems', function () {
+    cy.wait('@content');
+    cy.get('ul.desktop-menu a').contains('Just a page').click();
   });
 });
--- a/frontend/packages/volto-light-theme/news/848.feature
+++ b/frontend/packages/volto-light-theme/news/848.feature
@@ -1,0 +1,1 @@
+Fat menu now skips opening for top-level items that have no children — clicking the item navigates directly to its destination instead of opening an empty submenu. @ericof
--- a/frontend/packages/volto-light-theme/src/components/Navigation/Navigation.tsx
+++ b/frontend/packages/volto-light-theme/src/components/Navigation/Navigation.tsx
@@ -159,101 +159,109 @@ const Navigation = ({ pathname }: NavigationProps) => {
     >
       <div className={'computer large screen widescreen only'}>
         <ul className="desktop-menu">
-          {items.map((item, index) => (
-            <li key={item.url}>
-              {hasFatMenu ? (
-                <>
-                  <button
-                    onClick={() => openMenu(index)}
-                    className={cx('item', {
-                      active:
-                        desktopMenuOpen === index ||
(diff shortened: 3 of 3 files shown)


frontend/cypress/tests/main/nav.cy.js
```diff
@@ -51,4 +57,9 @@ context('Navigation Acceptance Tests', () => {
     cy.get('.subsubitem-wrapper').findByText('Level 3').click();
     cy.get('.documentFirstHeading').should('have.text', 'Level 3');
   });
+
+  it('Open destination page if no subitems', function () {
+    cy.wait('@content');
+    cy.get('ul.desktop-menu a').contains('Just a page').click();
```
davisagli, line 63: This test needs an actual assertion, otherwise it's just wasted time.

davisagli (changes requested):
The implementation looks ok.

The test is not a real test yet.

We should get @tisto's blessing on the UX change.

frontend/cypress/tests/main/nav.cy.js
ericof: @davisagli Fixed (I hope)

tisto:
@ericof, if navigation items look the same but their actions are different, we violate one of the most fundamental UX principle there is. We discussed this matter with clients multiple times and there are different solutions to the problem of navigation trees that are not filled. However, these solutions are more complex that what you implemented here. I am afraid we can not merge this PR and we need to discuss the matter more in depth before moving on.

davisagli:
@tisto What are the more complex solutions?

ericof:
> @ericof, if navigation items look the same but their actions are different, we violate one of the most fundamental UX principle there is

Hey @tisto, totally hear the concern about consistency — it's a real principle and worth taking seriously. I dug into the research a bit, though, and I think the picture is more nuanced than "same-looking items must behave the same." A few things worth sharing.

**The principle is about signifiers, not uniform behavior**. Nielsen Norman Group — literally the source of the heuristic — actively endorses mixed navigation in their [Menu-Design Checklist](https://www.nngroup.com/articles/menu-design/). They use Stanford University as the ✅ example: a caret icon distinguishes items that expand from items that go straight to a page. They note that "users may not even realize that there is a submenu if the category name doesn't look clickable" — which is the real problem the heuristic is trying to prevent. The fix is signifying the difference, not erasing it.

Personally,  I think the hover effect we have is the clear differentiation between the expandable and the clickable effect.

Anyway, if you don't think this could be merged, I'm fine with it, but I will open another PR to refactor the navigation component into smaller parts that could be easily overridden  -- as of today we need to customize the entire component just to add the checks I have added -- because every project I use VLT this is the first thing clients ask me to fix.

tisto:
@ericof I understand your concerns and I agree that this is a problem for sites that do have nav entries with no subitems. 

A fat menu that opens on click is a good fit for large websites that have the first three levels of the nav tree filled. If that is not the case, a fat menu that opens on click is just not the right solution for the given content.

We can add an option to open the fat menu on hover, as we did for fz-juelich.de. Then the UX is consistent, and the users can just click on a nav item to reach the page. However, we still need to open the fat menu on hover for items that do not have subelements.

Both options are valid and provide a consistent UX. If you want a different behavior per nav item (depending if they have subitems or not), you have to clearly indicate this as mentioned by:

https://www.nngroup.com/articles/menu-design/

as part 12. "Clearly Signify Submenus with a Caret or Arrow Icon".

Another option is to use classic submenus instead of a fat menu that indicates if a nav item has subitems (e.g. by a caret or arrow icon). However, we would have to be careful not to fall into:

14. Avoid Multilevel Cascading Menus; Use Mega Menus or Landing Pages Instead

https://www.nngroup.com/articles/menu-design/

Apart from that particular argument. We follow strict principles when working on VLT. We start with UX research and prototypes, then move to create and document the UI, and after we reached an agreement, we start to implement the solution. Our process includes consulting with UX experts, designers, and developers. If we want to maintain the quality of VLT that we currently have, we have to stick to this process. There is no shortcut.

tisto:
FYI: I would like to continue our discussion here: https://github.com/kitconcept/volto-light-theme/issues/456

tisto:
@ericof stripe.com has a fat menu with different behaviors per nav item. The key here is that the different behavior is clearly indicated by an arrow (if there are subitems, no arrow if there is no subitem):

Bildschirmfoto 2026-05-03 um 20 28 25

ericof:
Closing in favor of #904

PR #853, opened 2026-05-07
Load site customization fields on error views

davisagli:
@ericof @sneridagh This is my new solution for getting the site customization fields on error views, without needing to make a separate request (which caused a flash while the wrong content was in the redux store). Thoughts?

# Summary

- In the backend: override the exception view to add `['@components']['inherit']` from the closest visible context
- In the frontend: use a reducer to put the data from the exception view into `errorContext` in the redux store, and change `useLiveData` to fall back to that

### DIFF
--- a/frontend/packages/volto-light-theme/src/helpers/useLiveData.ts
+++ b/frontend/packages/volto-light-theme/src/helpers/useLiveData.ts
@@ -6,21 +6,27 @@ type FormState = {
   content: {
     data: Content;
   };
+  errorContext: Content;
   form: {
     global: Content;
   };
 };
 
 export function useLiveData<T>(
-  content: Content,
+  bbbContent: Content,
   behavior: string | undefined,
   field: string,
 ) {
+  const content = useSelector((state: FormState) => state.content?.data);
+  const errorContext = useSelector((state: FormState) => state.errorContext);
+  const context = content ?? errorContext;
+
   const location = useLocation();
   const addMode = location?.pathname?.endsWith('/add');
+
   const current = behavior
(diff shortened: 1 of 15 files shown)


frontend/packages/volto-light-theme/src/helpers/useLiveData.ts
```diff
@@ -6,21 +6,27 @@ type FormState = {
...
   };
+  errorContext: Content;
   form: {
     global: Content;
   };
 };
 
 export function useLiveData<T>(
-  content: Content,
+  bbbContent: Content,
```
davisagli: I think we can get rid of this argument (we get it from the redux store with getSelector instead), but that requires updating a lot of call sites, so I'll wait until I get feedback on the overall idea.

sneridagh: I can't recall if there was an underlying reason because of us passing down always the content... 
@iFlameing could you recall? I will take a look at the current shape of the instances.

sneridagh (approved):
Overall, I see it ok!

frontend/packages/volto-light-theme/src/helpers/useLiveData.ts
davisagli: @sneridagh I already forget why I changed it to use the redux state. So probably we could leave this as is.

davisagli:
@sneridagh Can you help me figure out why this broke one of the visual regression tests?

sneridagh:
@davisagli you had to update the branch, someone else updated the screenshots

sneridagh (approved):
LGTM

PR #861, opened 2026-05-15
Added the VR test for the controlpanel and the login script for that

TimoBroeskamp:
I added the Visual regression test and the login script that you need for the VR test to work.

When this PR gets merged we need to update the VR tests so that they are up to date with added test.


TimoBroeskamp:
The only thing failing is the visual regression test that I added which makes sense because I added it in this PR and we dont have a screenshot to compare.

PR #865, opened 2026-05-26
Polyglot translations for non-existing languages

erral:
This is the attempt to get the translations from Polyglot project by @mpalomaki

I will split the update in several PRs.

This PR contains only translations of those languages that currently VLT *does not have*.

I will create separate PRs for each of the language that VLT does have.


danalvrz:
@erral this is awesome! thank you, and also thanks to @mpalomaki!

sneridagh (approved):

PR #866, opened 2026-05-26
i18n: Update eu translations from Polyglot

erral:
Like #865 for eu

@mpalomaki


erral:
I reviewed the translations and are OK.

sneridagh (approved):

PR #869, opened 2026-05-26
i18n: Update pt_BR translations from Polyglot

erral:
Like #865 for pt-br

@mpalomaki


erral:
No new translations here. So closing.

PR #874, opened 2026-06-01
Fix last item of a listing with no pagination padding is broken

iRohitSingh:
Issues Link: https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-intranet/-/work_items/391

### DIFF
--- a/frontend/cypress/tests/core/basic/search.js
+++ b/frontend/cypress/tests/core/basic/search.js
@@ -121,6 +121,6 @@ describe('Search', () => {
 
     // then the first link must be Colorless
     cy.get('button[name="effective"]').click();
-    cy.get('.summary.url:first').should('have.text', 'Color');
+    cy.get('.summary.url:first').should('have.text', 'Colorless');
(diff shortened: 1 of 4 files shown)


iFlameing (approved):
LGTM!

davisagli:
@iRohitSingh @iFlameing Look at the CI checks -- the visual regression test images need to be updated /cc @sneridagh

sneridagh:
@danalvrz @iFlameing @iRohitSingh Given the regression tests, I think the selector had sense for non-grids (summaries). So we have to fix the grids, and leave alone the summaries. Can you take a look at that? 

Question: did you take a look at the tests locally? because it makes total sense after looking at them:

image


That's precisely why we set them up, to make sense of this things. Download, extract and:

`pnpm exec playwright show-trace ...trace.zip`

sneridagh:
@danalvrz final review?

danalvrz (approved):
looks good!

PR #876, opened 2026-06-04
Fix Size widget position

iRohitSingh:
Issues Link: https://gitlab.kitconcept.io/kitconcept/distribution-kitconcept-intranet/-/work_items/396

ScreenShot:

Screenshot

### DIFF
--- a/frontend/packages/volto-light-theme/news/876.bugfix
+++ b/frontend/packages/volto-light-theme/news/876.bugfix
@@ -1,0 +1,1 @@
+Fix Size widget position @iRohitSingh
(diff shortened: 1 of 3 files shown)


sneridagh:
@iRohitSingh the selector is wrong... I have no clue how this was fixed.

image

sneridagh:
@iRohitSingh Please see https://github.com/plone/volto/pull/8033. I'd have liked that you did this research and be more curious about what really caused this, and find out what I found. Let's wait for Wagner to answer.

iRohitSingh:
@sneridagh There is one problem with this extra wrapper div, but since it is used everywhere, I thought it was not the issue because the other widgets are aligned correctly. So, I investigated further and found that it is even leaking from its parent container.

The main problem is in ImageSizeWidget, where we are using Semantic UI Grid components. Each Semantic UI Grid component has negative margins on all sides. That's why it is moving up due to the negative margins.

Negative_Margin

sneridagh:
This was fixed in Volto 19.1.2. There was a regression in the widgets due to a a11y addition.

PR #882, opened 2026-06-10
Fix advanced facet toggle not working in Search block.

iFlameing:


kittauri:
Works locally. LGTM

kittauri:
Note: Maybe use the opportunity to get rid of semantic-ui-react Grid ?

sneridagh:
@kittauri We have the policy that all shadows that are not main blocks (header, footer, etc) we will continue to use SemanticUI for simplicity (and if in case that this shadow can go away at some point, because the change made it to core), also for comparisons on "what this shadow does" occasions.

iFlameing:
@danalvrz can you please approve this we can merge this since @kittauri already made the comment. He can't approve because he doesn't have permission.

danalvrz (approved):
LGTM

PR #891, opened 2026-06-12
adjusted css to match DOM

TimoBroeskamp:
Changed a css selector to apply the css accordingly.


TimoBroeskamp:
@danalvrz you made this change three months ago. Do you remember why?

danalvrz:
@TimoBroeskamp Just to clarify, the file `_search-page.scss` is meant for the search results page coming from our Solr service (as mentioned in the comment on the top). We need to find a way of updating the default Volto search results page without breaking the Solr implementation.

TimoBroeskamp:
Before merging this I need to know how to test that with solr active. I'll put this on not ready for that reason.

PR #892, opened 2026-06-12
The “Link Content Type URL” was not resolved correctly

iFlameing:
@sneridagh I really wanted to improve the link of Fat Menu to use UniversalLink instead of react-router-dom because our UniversalLink handle all the case.


sneridagh (approved):

sneridagh:
@iFlameing the tests are failing. Also the `isActive` method is ok? can you write a test for it, and edge cases, just in case? :)

iFlameing:
> @iFlameing the tests are failing. Also the `isActive` method is ok? can you write a test for it, and edge cases, just in case? :)

I don't know why vrt test are failing, first time it fails on vertical spacing, second time on homepage. Let's see this time. 
`main` is already merge. I added the unit test for isActive function.

PR #899, opened 2026-06-26
Migrate Image block size to `size:noprefix` style field

danalvrz:


sneridagh (approved):
LGTM. @iFlameing @iRohitSingh can you double check and check it locally? Specially with old data, (coming from main) and then switch to this branch). I could not test it locally and I'd love a second pair of eyes on it. Thx!!

iFlameing (approved):
@sneridagh @danalvrz I tested locally and it worked.

PR #905, opened 2026-07-16
Let projects choose structural components through the registry

### LINKED ISSUE #904: Let projects choose structural components through the registry
## Problem

Replacing a structural component — the header, the navigation, the footer — currently means shadowing it through `customizations/`.

That works, but it is a blunt instrument:

- **It is global and all-or-nothing.** A shadow replaces the component for the whole build. Two variants cannot coexist, so there is no way to ship an alternative navigation and let a site opt into it.
- **It is invisible.** Nothing in the configuration says a component has been replaced. The only way to find out is to notice the file under `customizations/`, and a shadow that no longer matches an upstream path fails silently — the alias simply resolves for nobody and the upstream component keeps rendering.
- **It couples a project to internal structure.** A shadow is bound to the exact module path. When the theme moves a file, the customization stops applying, again silently.
- **It resists composition.** Two add-ons that both want to influence the header cannot; the last shadow wins.

The theme already registers slots and blocks through the registry, where a project can add, replace or reorder by name. Structural components are the outlier: they are the pieces most likely to need replacing, and the only ones with no configuration-level seam.

## Proposal

Register the structural components as **named utilities** and let settings pick which name renders.

The theme registers its own implementations under a stable name (`vlt`), one per component type:

```js
config.registerUtility({ name: 'vlt', type: 'header', method: Header });
config.registerUtility({ name: 'vlt', type: 'navigation', method: Navigation });
config.registerUtility({ name: 'vlt', type: 'mobileNavigation', method: MobileNavigation });
config.registerUtility({ name: 'vlt', type: 'footer', method: Footer });
```

The selection becomes an ordinary setting:

```js
config.settings.vlt = {
  components: {
    header: 'vlt',
    navigation: 'vlt',
    mobileNavigation: 'vlt',
    footer: 'vlt',
  },
};
```

A project then registers its own implementation under its own name and flips one setting — no shadowing, and both implementations stay available:

```js
config.registerUtility({ name: 'acme', type: 'navigation', method: AcmeNavigation });
config.settings.vlt.components.navigation = 'acme';
```

Defaults keep today's behaviour, so this is backwards compatible.

## What needs to change

**1. Register the components as utilities** — one `registerUtility` per structural component, all under `name: 'vlt'`, with `type` naming the role (`header`, `navigation`, `mobileNavigation`, `footer`).

**2. Add a resolver helper** that turns a type into a component:

```js
const DEFAULT = 'vlt';

export function getVLTComponent(type) {
  const name = config.settings.vlt?.components?.[type] || DEFAULT;
  const component = config.getUtility({ name, type }).method;
  return component || config.getUtility({ name: DEFAULT, type }).method;
}
```

The fallback matters and it is safe: `@plone/registry`'s `getUtility` returns `{}` on a miss rather than throwing, so a setting that names an unregistered component degrades to the theme's own rather than crashing the render. Worth an explicit test, since it is the behaviour the whole design leans on.

**3. Make the theme's own components resolve through the lookup.** The theme customizes Volto's `Header` and `Footer`; those customizations should render whatever `getVLTComponent` returns rather than importing a component directly. That indirection is the actual seam — without it, settings have nothing to act on.

**4. Type the settings** — augment `SettingsConfig` so `settings.vlt.components` is typed, and prefer `keyof` over a bare `string` for the component type. A typo in a type name otherwise compiles and surfaces as *"Element type is invalid"* at render.

**5. Document it** — a short section showing "register your own, flip the setting" is what makes this discoverable, which is most of the point.

## Optional: introspection

A helper listing the implementations registered for each type, and which one is currently selected, makes the mechanism inspectable and gives control panels or a demo something to build on:

```js
listComponents()
// { navigation: [ { name: 'vlt', selected: false }, { name: 'acme', selected: true } ], ... }
```

Two details for whoever writes it:

- **The names are the registry's object keys.** `registerUtility` stores `{ method }` only — `getUtilities({ type })` hands back the methods with the names dropped, so the names have to come from the keys of `config._data.utilities[type]`.
- **Skip dependency-registered utilities.** Those are keyed with a `|<dependencies>` prefix. Since the resolver looks utilities up without dependencies, it can never select one, so listing them would offer a choice that cannot be taken.

Consider making `selected` mirror the resolver's fallback rather than echoing the raw setting: if the configured name is not registered, the theme's own component is what actually renders, and a listing that says otherwise disagrees with the screen.

## Acceptance criteria

- [ ] Header, navigation, mobile navigation and footer are each registered as a utility under `vlt`
- [ ] `settings.vlt.components` selects which registered name renders for each type
- [ ] A project can register its own component and select it without any `customizations/` shadow
- [ ] An unregistered name falls back to the theme's own component instead of failing to render
- [ ] Defaults reproduce today's behaviour exactly
- [ ] The settings are typed, and an invalid component type is a compile error
- [ ] Tests cover the lookup, the fallback, and the no-settings case
- [ ] Documented with a worked "swap the navigation" example

ericof:
## Summary

Structural components — the header, navigation, footer, and several more — could previously only be replaced by shadowing them through `customizations/`. That is global, invisible in configuration, and coupled to internal module paths.

This PR registers each structural component as a `@plone/registry` utility under the name `vlt`, and resolves it through `config.settings.vlt.components` at render time. A project can register its own implementation and select it with a single setting, no shadowing required, with both implementations staying available.

Closes #904

## Swappable components

Nine roles can be selected via `config.settings.vlt.components`:

`breadcrumbs`, `footer`, `header`, `languageSelector`, `logo`, `mobileNavigation`, `navigation`, `searchWidget`, `tags`.

## How a project uses it

```js
// Register your own under your own name.
config.registerUtility({ name: 'acme', type: 'navigation', method: AcmeNavigation });

// Select it.
config.settings.vlt.components.navigation = 'acme';
```

## What changed

- **`config/components.ts`** — registers every structural component as a `vlt` utility. The registration map is keyed by `ComponentSlot`, so TypeScript requires an entry whenever a slot is added to the settings type.
- **`helpers/settings.ts`** — `getVLTComponent(type)` resolver with a safe fallback to the theme's own component when a configured name is not registered.
- **Header/Footer and the individual component customizations** — now resolve through `getVLTComponent` instead of re-exporting a fixed component. This indirection is the seam the setting acts on.
- **`config/settings.ts` / `types.d.ts`** — typed `VLTSettings` augmenting `SettingsConfig`, with defaults that reproduce today's behavior exactly.

## Backwards compatibility

Defaults select `vlt` for every role, so an existing site renders identically. An unregistered name falls back to the theme's own component rather than failing to render.

## Testing

- New unit tests cover the resolver: correct selection, fallback on an unregistered name, and the no-settings case.
- Full frontend unit suite passes (88/88).

## Documentation

New how-to guide: **Swap structural components**, with a worked "swap the navigation" example.

----
📚 Documentation preview 📚: https://volto-light-theme--905.org.readthedocs.build/

### DIFF
--- a/docs/how-to-guides/swap-structural-components.md
+++ b/docs/how-to-guides/swap-structural-components.md
@@ -1,0 +1,117 @@
+---
+myst:
+  html_meta:
+    "description": "Swap structural components in Volto Light Theme through the registry"
+    "property=og:description": "Swap structural components in Volto Light Theme through the registry"
+    "property=og:title": "Swap structural components"
+    "keywords": "Volto Light Theme, registry, utility, header, navigation, footer, customization"
+---
+
+# Swap structural components
+
+Volto Light Theme (VLT) renders its structural components — the header, navigation, footer, and a few more — through the component registry from `@plone/registry`.
+This lets a project replace any of them with its own implementation by registering a utility and flipping a setting, instead of shadowing the component through {file}`customizations`.
+
+Compared to shadowing, this approach is:
+
+-   **Explicit.** The active component is named in configuration, not hidden in a file under {file}`customizations`.
+-   **Composable.** Several implementations can coexist under different names; a project selects which one renders.
+-   **Decoupled.** A project binds to a stable name, not to an internal module path.
+
+## How it works
+
+VLT registers its own implementation of each structural component as a utility under the name `vlt`, with a `type` that names the role.
+
+```js
+config.registerUtility({ name: 'vlt', type: 'navigation', method: Navigation });
+```
+
+A setting then selects which registered name renders for each role.
+
+```js
+config.settings.vlt = {
+  components: {
+    breadcrumbs: 'vlt',
+    footer: 'vlt',
+    header: 'vlt',
+    languageSelector: 'vlt',
+    logo: 'vlt',
+    mobileNavigation: 'vlt',
+    navigation: 'vlt',
+    searchWidget: 'vlt',
+    tags: 'vlt',
+  },
+};
+```
+
+These defaults reproduce the theme's standard behavior, so enabling nothing changes nothing.
+
+## Available components
+
+The following roles can be swapped. The value is the `type` you register against and the key you set under `config.settings.vlt.components`.
+
+| Setting            | Role                                            |
+| ------------------ | ----------------------------------------------- |
+| `breadcrumbs`      | The breadcrumbs trail                           |
+| `footer`           | The site footer                                 |
+| `header`           | The site header                                 |
+| `languageSelector` | The language selector in the header             |
+| `logo`             | The site logo                                   |
+| `mobileNavigation` | The mobile (hamburger) navigation               |
+| `navigation`       | The main desktop navigation                     |
+| `searchWidget`     | The header search widget                        |
+| `tags`             | The tags shown on content                       |
+
+## Swap the navigation
+
+Suppose your project add-on `@acme/volto` ships its own navigation and you want the site to use it.
+
+### 1. Register your component
+
+In your add-on's configuration, register your navigation under your own name and the `navigation` type.
+
+```js
+import AcmeNavigation from './components/AcmeNavigation';
+
+export default function applyConfig(config) {
+  config.registerUtility({
+    name: 'acme',
+    type: 'navigation',
+    method: AcmeNavigation,
+  });
+
+  return config;
+}
+```
+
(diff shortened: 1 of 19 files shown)


sneridagh (approved):
LGTM!! Great addition!!

davisagli (commented):
@ericof @sneridagh Making these configurable in the registry is nice. But I don't really understand why you added a new `vlt` utility type instead of using the existing ability to configure slots.

PR #908, opened 2026-07-24
Fix regression links proxy object

Tishasoumya-02:
After testing https://github.com/kitconcept/kitconcept.intranet/pull/407 there were some regressions with the links, fix in this PR


danalvrz (approved):
LGTM, but please take a look at the failing VRT before we can merge it. It seems like the internal link example in /content-types/event is showing the external link icon.

Tishasoumya-02:
All tests passing now!

PR #910, opened 2026-08-10
Enhance the banner block's `blockWidth:noprefix` schema 

Tishasoumya-02:


frontend/packages/volto-light-theme/src/components/Blocks/Banner/schema.js
```diff
@@ -0,0 +1,13 @@
+import { defaultStylingSchema } from '../schema';
+
+export const BannerStylingSchema = ({ schema, formData, intl }) => {
+  defaultStylingSchema({ schema, formData, intl });
+  const styleProps = schema.properties.styles?.schema?.properties;
+
+  if (styleProps?.['blockWidth:noprefix']) {
```
iFlameing, line 7: @Tishasoumya-02 this is totally wrong see my comment in banner block repo to properly fix it.

iFlameing:
@danalvrz @sneridagh i want to merge this because it follow the same approach as other addons. I don't satisfied with solution but first let's merge this and then we will fix all the addons using proper fix.

https://github.com/kitconcept/volto-light-theme/blob/main/frontend/packages/volto-light-theme/src/components/Blocks/Button/schema.js#L13

https://github.com/kitconcept/volto-light-theme/blob/main/frontend/packages/volto-light-theme/src/components/Blocks/Separator/schema.js#L13

iFlameing (approved):
Similar to others.

sneridagh:
We discussed and decided to go for the change in banner add-on itself. Let's fix CI there, make sure that VLT is in the env too.

iFlameing:
closing because we fix this in addon and release it also updated it here.

PR #918, opened 2026-09-03
Fix duplicated style classnames

danalvrz:


frontend/cypress/tests/core/blocks/blocks-teaser.js
```diff
@@ -111,6 +111,7 @@ context('Blocks Acceptance Tests', () => {
 
     cy.get('.documentFirstHeading').contains('Blue Orchids and Tulips');
     // THEN I can see the updated content in the teaser
+    cy.wait(500);
```
davisagli: I don't understand how a wait immediately before navigating is helpful

sneridagh:
@danalvrz As we discussed the other day... I think this is too much, I'd rather leave it as it is. Live with the duplicated classnames, maybe deprecate them in v9. What do you think? let's close this one?

danalvrz:
ok, let's wait a little longer before this change, closing the PR for now.

